PR #19, opened 2015-05-22
Fixes append_path.

trolleyman:
This fixes the bug where upon appending a path GTK complains and the context no longer works: "Gtk-WARNING **: drawing failure for widget `GtkDrawingArea': invalid value for an input cairo_status_t"

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -683,7 +683,7 @@ impl Context {
 
     pub fn append_path(&self, path: &Path){
         unsafe {
-            ffi::cairo_append_path(self.get_ptr(), transmute(path))
+            ffi::cairo_append_path(self.get_ptr(), transmute(path.get_ptr()))
         }
     }
 


GuillaumeGomez:
Nice one, thanks ! I'll merge once check is done.

src/context.rs
```diff
@@ -683,7 +683,7 @@ impl Context {
 
     pub fn append_path(&self, path: &Path){
         unsafe {
-            ffi::cairo_append_path(self.get_ptr(), transmute(path))
+            ffi::cairo_append_path(self.get_ptr(), transmute(path.get_ptr()))
```
gkoz: The transmute isn't needed anymore is it?

GuillaumeGomez: I didn't even see the transmute...

trolleyman: Oh yeah, probably not anymore

PR #24, opened 2015-06-08
Add documentation

GuillaumeGomez:
Still on progress.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -33,8 +33,6 @@ pub struct RectangleVec {
 }
 
 impl Drop for RectangleVec {
-    /// Unconditionally frees rectangle_list and all associated references. After this call,
-    /// the rectangle_list pointer must not be dereferenced.
     fn drop(&mut self) {
         unsafe {
             ffi::cairo_rectangle_list_destroy(self.ptr);
--- a/src/context.rs
+++ b/src/context.rs
@@ -67,7 +67,7 @@ impl Context {
         }
     }
 
-    /// Increases the reference count on cr by one. This prevents cr from being destroyed
+    /// Increases the reference count on self by one. This prevents self from being destroyed
     /// until a matching call to drop() is made.
     /// 
     /// The number of references to a cairo_t can be get using Context::get_reference_count().
@@ -84,7 +84,7 @@ impl Context {
     /// Creates a new Context with all graphics state parameters set to default values
     /// and with target as a target surface. The target surface should be constructed
     /// with a backend-specific function such as cairo_image_surface_create() (or any other
-    /// cairo_backend_surface_create() variant).
+    /// Context::backend_surface_create() variant).
     /// 
     /// This function references target , so you can immediately call Surface::drop() on
     /// it if you don't need to maintain a separate reference to it.
@@ -213,7 +213,7 @@ impl Context {
 
     //fn ffi::cairo_get_group_target (cr: *mut cairo_t) -> *mut cairo_surface_t;
 
-    /// Sets the source pattern within cr to an opaque color. This opaque color will then be
+    /// Sets the source pattern within self to an opaque color. This opaque color will then be
     /// used for any subsequent drawing operation until a new source pattern is set.
     /// 
     /// The color components are floating point numbers in the range 0 to 1.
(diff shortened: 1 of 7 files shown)


gkoz:
:+1: nice

GuillaumeGomez:
Well, it's just the beginning. A lot remains here.

GuillaumeGomez:
@gkoz: I think I'm done here. Can you just take a quick look before I merge please ?

src/context.rs
```diff
@@ -31,6 +33,8 @@ pub struct RectangleVec {
 }
 
 impl Drop for RectangleVec {
+    /// Unconditionally frees rectangle_list and all associated references. After this call,
+    /// the rectangle_list pointer must not be dereferenced.
     fn drop(&mut self) {
```
gkoz: I don't think documenting destructors makes sense.

src/context.rs
```diff
@@ -63,28 +67,49 @@ impl Context {
         }
     }
 
+    /// Increases the reference count on self by one. This prevents self from being destroyed
+    /// until a matching call to drop() is made.
+    /// 
+    /// The number of references to a cairo_t can be get using Context::get_reference_count().
```
gkoz: This probably wants to be `#[doc(hidden)]` or put onto the `glib::translate` rails.

src/context.rs
```diff
@@ -63,28 +67,49 @@ impl Context {
...
             Context(ffi::cairo_reference(self.get_ptr()))
         }
     }
 
-    pub fn ensure_status(&self){
+    pub fn ensure_status(&self) {
         self.status().ensure_valid();
     }
 
+    /// Creates a new Context with all graphics state parameters set to default values
```
gkoz: Don't we have to put identifiers and code snippets in back-quotes?

src/context.rs
```diff
@@ -101,7 +128,41 @@ impl Context {
...
+    /// Context::restore()).
+    /// 
+    /// By default the intermediate group will have a content type of Content::ColorAlpha.
+    /// Other content types can be chosen for the group by using
+    /// Context::push_group_with_content() instead.
+    /// 
+    /// As an example, here is how one might fill and stroke a path with translucence, but
+    /// without any portion of the fill being visible under the stroke:
+    /// 
+    /// ```
```
gkoz: We might want this to be "ignore" so `cargo test` won't barf on it. (https://doc.rust-lang.org/book/documentation.html#running-documentation-tests)

src/context.rs
GuillaumeGomez: I was wondering when documenting but you're right, with a step backward, it doesn't make sense.

src/context.rs
GuillaumeGomez: It's quite complicated since the cairo API is quite old now. I think a big work on updating and finishing this crate is needed. For now, I just hide get_ptr functions.

src/context.rs
GuillaumeGomez: I just did with code snippets. Is it really necessary for identifiers ?

src/context.rs
GuillaumeGomez: I thought it was just a code snippet ? I'm not sure it does test it...

src/context.rs
gkoz: It's just what everybody does. ;) I know it would be frustrating to do by hand. Might be possible to do with some regex incantations.

src/context.rs
gkoz: The tests will fail if any of the examples not marked as `ignore` don't build. If someone gets around to adding tests, they'll get a lot of failures. Someone adding tests may not be very likely though. =)

src/context.rs
GuillaumeGomez: Well, I'll have to go back on code then. :'(

src/context.rs
GuillaumeGomez: Erf, that's so annoying...

GuillaumeGomez:
It's ready (I added `ignore`). I'll update cairotest too.

gkoz:
My head hurts after thinking about the GObject type system internals =)

GuillaumeGomez:
No problem. ;-)

I saw the question about adding custom widget. Do you think we'll add something which will make it possible ? For me, it seems very unlikely. We didn't take this path from the beginning after all....

gkoz:
It could be an interesting challenge and maybe even Rust-C interoperability case study. But I haven't seen a strong enough need for this yet.

It's fair to want to have custom widget types. Take that `Clock` example. You have some internal fields and you want to access them from a signal handler. Currently widget signal handlers take `Widget` as a receiver. If they could take that custom type instead, they'd avoid some `Rc` dances. It could also be considered lunacy to make a sandwich of glib-allocated buffers that hold Rust data that is accessed by Rust code called from C callbacks. Too bad that absent stable Rust ABI we pretty much can't have a native toolkit.

GuillaumeGomez:
Hum... It'd be interesting to give some feedback to rust core developers in order to make it possible if not yet in nightly.

PR #32, opened 2015-07-19
Replace features with cfg and autodetection

gkoz:


GuillaumeGomez:
Thanks ! I wait for checks before merging.

gkoz:
Yeah there's an ICE, I'll need to update the Rust version in appveyor first.

GuillaumeGomez:
I have an ICE on gtk. It must be the bad week. I'll wait for your update then. ^^

GuillaumeGomez:
And the problem is also on travis. Congrats ! haha

gkoz:
Apparently `rustc` doesn't like bare `--cfg` and cargo doesn't bother to check.

GuillaumeGomez:
You're using a deprecated item in `build.rs`.

gkoz:
`join` hasn't made it into stable yet so...

GuillaumeGomez:
erf... So annoying. Let's merge it for now. Thanks a lot for your work !

PR #33, opened 2015-07-29
Add partial Surface and ImageSurface bindings

gkoz:


GuillaumeGomez:
Very big work, thanks a lot! Everything's fine, I merge.

PR #37, opened 2015-09-06
Org rename

gkoz:
Forgot this one

PR #38, opened 2015-09-06
Revert the wrong "Org rename"

gkoz:
Reverts gtk-rs/cairo#37


GuillaumeGomez:
Why not just changing Rgtk to Gtk-rs instead of reverting ?

gkoz:
I have the renaming incantations in my shell history from rust-gnome, it's more reliable than to do from scratch

GuillaumeGomez:
As you wish.

PR #40, opened 2015-09-06
Org rename

gkoz:


gkoz:
Let's see if travis will be green

GuillaumeGomez:
It should be.

GuillaumeGomez:
We're good !

PR #46, opened 2015-10-17
Update authors

gkoz:


GuillaumeGomez:
We should add an autotool for this, it's a bit annoying to add it ourselves (thanks for doing it btw ! ^^).

gkoz:
I have a local repo that tracks the committers and then it's only takes editing the file once and the rest is just a bunch of shell commands.

PR #60, opened 2016-02-15
PNG support

vojtechkral:
This depends on `CAIRO_HAS_PNG_FUNCTIONS`, is that a problem?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ name = "cairo"
 
 [features]
 "1.12" = ["cairo-sys-rs/1.12"]
+"png" = [ "cairo-sys-rs/png" ]
 
 [dependencies.cairo-sys-rs]
 path = "./cairo-sys-rs"
--- a/src/image_surface_png.rs
+++ b/src/image_surface_png.rs
@@ -1,0 +1,92 @@
+// Copyright 2016, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use std::slice;
+use std::io::{Read, Write, Error};
+use std::process;
+use std::thread;
+
+use libc::{c_void, c_uint};
+
+use glib::translate::*;
+use ffi;
+use ffi::enums::Status;
+use error::IoError;
+use ImageSurface;
+
+
+struct CallbackGuard;
+
+impl Drop for CallbackGuard {
+    fn drop(&mut self) {
+        if thread::panicking() {
+            process::exit(101);
+        }
+    }
+}
+
+struct ReadEnv<R: Read> {
+    reader: R,
+    error: Option<Error>,
+}
+
+unsafe extern "C" fn read_func<R: Read>(closure: *mut c_void, data: *mut u8, len: c_uint) -> Status {
+    let _cbguard = CallbackGuard;
+    let read_env: &mut ReadEnv<R> = &mut *(closure as *mut ReadEnv<R>);
+    let mut buffer = slice::from_raw_parts_mut(data, len as usize);
+    match read_env.reader.read_exact(buffer) {
+        Ok(()) => Status::Success,
+        Err(error) => {
+            read_env.error = Some(error);
+            Status::ReadError
+        },
+    }
+}
+
+struct WriteEnv<W: Write> {
+    writer: W,
+    error: Option<Error>,
+}
+
+unsafe extern "C" fn write_func<W: Write>(closure: *mut c_void, data: *mut u8, len: c_uint) -> Status {
+    let _cbguard = CallbackGuard;
+    let write_env: &mut WriteEnv<W> = &mut *(closure as *mut WriteEnv<W>);
+    let buffer = slice::from_raw_parts(data, len as usize);
+    match write_env.writer.write_all(buffer) {
+        Ok(()) => Status::Success,
+        Err(error) => {
+            write_env.error = Some(error);
+            Status::WriteError
+        },
+    }
+}
+
+
+impl ImageSurface {
+    pub fn create_from_png<R: Read>(stream: R) -> Result<ImageSurface, IoError> {
+        let mut env = ReadEnv{ reader: stream, error: None };
+        let surface: ImageSurface = unsafe { from_glib_full(ffi::cairo_image_surface_create_from_png_stream(
+            Some(read_func::<R>), &mut env as *mut ReadEnv<R> as *mut c_void)) };
+        match env.error {
+            None => match surface.as_ref().status() {   // The surface migth still be "nil" if the error occured in Cairo
+                Status::Success => Ok(surface),
+                st => Err(IoError::Cairo(st)),
+            },
+            Some(err) => Err(IoError::Io(err)),
+        }
+    }
+
+    pub fn write_to_png<W: Write>(&self, stream: W) -> Result<(), IoError> {
+        let mut env = WriteEnv{ writer: stream, error: None };
+        let status = unsafe { ffi::cairo_surface_write_to_png_stream(self.to_glib_none().0,
+            Some(write_func::<W>), &mut env as *mut WriteEnv<W> as *mut c_void) };
+        match env.error {
+            None => match status {
+                Status::Success => Ok(()),
+                st => Err(IoError::Cairo(st)),
+            },
+            Some(err) => Err(IoError::Io(err)),
+        }
+    }
+}
--- a/cairo-sys-rs/Cargo.toml
+++ b/cairo-sys-rs/Cargo.toml
@@ -19,6 +19,7 @@ name = "cairo_sys"
 [features]
 "1.12" = []
 "1.14" = ["1.12"]
+"png" = []
 
 [dependencies]
 libc = "0.2"
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright 2013-2015, The Gtk-rs Project Developers.
+// Copyright 2013-2016, The Gtk-rs Project Developers.
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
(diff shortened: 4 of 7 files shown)


gkoz:
Uh, do we have an idiomatic error handling solution in cairo?

Looks like they [recommend](http://cairographics.org/manual/bindings-streams.html) we bind the `_stream` functions in an idiomatic way (`fn create_from_png<R: Read>(r: R) -> Result<ImageSurface, IoError>`). `IoError` might for now be something like `enum { Cairo(Status), Io(io::Error) }`

As to `CAIRO_HAS_PNG_FUNCTIONS`, we have cargo features again so it seems natural to put this behind a feature and investigate if it can be enabled by default.

vojtechkral:
Ah, yes, binding the `_stream` functions is probably a better idea... I'm not sure I'd be able to properly cast the stream object to `void*` for the closure though.

I've noticed gtk-rs projects now generate bindings using GIR, which I don't have much idea about, but this isn't the case with cairo is it?

GuillaumeGomez:
No it's not unfortunately.

gkoz:
> I'm not sure I'd be able to properly cast the stream object to void* for the closure though.

Full-blown closures a hardly needed here but it's possible to write a generic callback e.g.

```rust
struct ReadEnv<R: Read> {
    reader: R,
    error: Option<io::Error>,
}

extern fn read_func<R: Read>(closure: *mut c_void, data: *mut u8, len: c_uint) -> Status {
    CallbackGuard; // "borrow" this from glib or gtk crate
    unsafe {
        let read_env: &mut ReadEnv<R> = &mut *(closure as *mut ReadEnv);
        ...
```

You might encounter some pushback from the borrow checker but the lifetime of `R` should not be a practical problem here.

vojtechkral:
Ah, shoot, can't use `#![feature(read_exact)]` on stable...

Cargo.toml
```diff
@@ -18,6 +18,7 @@ name = "cairo"
 
 [features]
 "1.12" = ["cairo-sys-rs/1.12"]
+"png-functions" = [ "cairo-sys-rs/png-functions" ]
```
gkoz: I'd prefer just `png`.

src/image_surface.rs
```diff
@@ -1,19 +1,85 @@
...
 use ffi::enums::{
+    Status,
     Format,
     SurfaceType,
 };
 
 use surface::{Surface, SurfaceExt, SurfacePriv};
 
+#[cfg(feature = "png-functions")]
+mod png_functions {
```
gkoz: It seems tidier to put all png stuff in a separate `image_surface_png` module

src/image_surface.rs
```diff
@@ -44,6 +110,31 @@ impl ImageSurface {
...
+    pub fn create_from_png_stream<R: Read>(stream: R) -> io::Result<ImageSurface> {
+        let mut env = png_functions::ReadEnv{ reader: stream, error: None };
+        let surface: ImageSurface = unsafe { from_glib_full(ffi::cairo_image_surface_create_from_png_stream(
+            transmute(png_functions::read_func::<R>), transmute(&mut env))) };
+        match env.error {
+            None => match surface.0.status() {   // The surface migth still be "nil" if the error occured in Cairo
+                Status::Success => Ok(surface),
+                Status::ReadError => Err(io::Error::new(io::ErrorKind::InvalidData, "Invalid PNG data.")),
+                Status::NoMemory => Err(io::Error::new(io::ErrorKind::InvalidData, "Could not allocate enough memory.")),
+                _ => Err(io::Error::new(io::ErrorKind::InvalidData, "Unknown error.")),
```
gkoz: It doesn't seem right to just hide `Status`. Clearly `io::Error` can't reflect some failure modes, notably `NoMemory`.

Until we have a good error handling story an `IoError` enum I suggested earlier seems preferable.

src/image_surface.rs
```diff
@@ -44,6 +110,31 @@ impl ImageSurface {
         }
     }
 
+    #[cfg(feature = "png-functions")]
+    pub fn create_from_png_stream<R: Read>(stream: R) -> io::Result<ImageSurface> {
```
gkoz: I believe the best approximation of [this guideline](http://cairographics.org/manual/bindings-overloading.html) would be to avoid the `_stream` suffix as this is the more generic version. Optionally there could be variants that take `P: AsRef<Path>`.

src/image_surface.rs
```diff
@@ -1,19 +1,85 @@
...
+
+    pub unsafe extern "C" fn write_func<W: Write>(closure: *mut c_void, data: *mut u8, len: c_uint) -> Status {
+        let _cbguard = CallbackGuard;
+        let write_env: &mut WriteEnv<W> = &mut *(closure as *mut WriteEnv<W>);
+        let buffer = slice::from_raw_parts(data, len as usize);
+        match write_env.writer.write_all(buffer) {
+            Ok(()) => Status::Success,
+            Err(error) => {
+                write_env.error = Some(error);
+                Status::ReadError
```
gkoz: I'd feel safer with `WriteError` here ;)

vojtechkral: Damn :)

src/image_surface.rs
vojtechkral: Ok, sure, I can do that...

The reason I haven't done this right away is that, at least as of current Cairo implementation, the error return codes seem to be pretty much random when Cairo is fed an invalid PNG. It will either return a `ReadError` or `NoMemory`, presumably depending on the kind of corruption of the PNG data, but we can't really tell whether we're actually out of memory or whether the PNG was just corrupted...

gkoz: Eh, they do seem to take some shortucts but then also return a number of statuses that make sense.

src/image_surface.rs
vojtechkral: Hm, I'm not sure I follow... I can just omit the `Path` version? Or, if not, what should it be named?

gkoz: Personally I see little value in providing `Path` versions that would amount to e.g. 
```rust
create_from_png(try!(File::open(path)))
```

src/image_surface.rs
```diff
@@ -15,7 +15,7 @@ use ffi::enums::{
 use surface::{Surface, SurfaceExt, SurfacePriv};
 
 #[derive(Debug)]
-pub struct ImageSurface(Surface);
+pub struct ImageSurface(pub Surface);
```
gkoz: This should not be necessary. If it is there's a bug somewhere.

src/image_surface_png.rs
```diff
@@ -0,0 +1,110 @@
...
+#[derive(Debug)]
+pub enum PngError { Cairo(Status), Io(Error) }
+
+pub trait ImageSurfacePng: Sized {
+    fn create_from_png<R: Read>(stream: R) -> Result<Self, PngError>;
+    fn write_to_png<W: Write>(&self, stream: W) -> io::Result<()>;
+}
+
+impl ImageSurfacePng for ImageSurface {
+    fn create_from_png<R: Read>(stream: R) -> Result<ImageSurface, PngError> {
```
gkoz: You can just `impl ImageSurface` without introducing a trait.

src/image_surface_png.rs
```diff
@@ -0,0 +1,110 @@
...
+        Err(error) => {
+            write_env.error = Some(error);
+            Status::WriteError
+        },
+    }
+}
+
+
+#[derive(Debug)]
+pub enum PngError { Cairo(Status), Io(Error) }
```
gkoz: There's nothing tying this enum to png. I'd rather it had a more generic name e.g. `IoError` and live in its own module, say `error`.

src/image_surface_png.rs
```diff
@@ -0,0 +1,110 @@
...
+            Ok(n) => { let tmp = buffer; buffer = &mut tmp[n..]; }
+            Err(ref e) if e.kind() == ErrorKind::Interrupted => {}
+            Err(e) => {
+                read_env.error = Some(e);
+                return Status::ReadError;
+            }
+        }
+    }
+    if !buffer.is_empty() {
+        read_env.error = Some(Error::new(ErrorKind::Other, "failed to fill whole buffer"));
```
gkoz: Would `ErrorKind::UnexpectedEof` be appropriate here?

gkoz: Hey, `read_exact` is stable in 1.6. We can use it.

src/image_surface_png.rs
```diff
@@ -0,0 +1,110 @@
...
+pub trait ImageSurfacePng: Sized {
+    fn create_from_png<R: Read>(stream: R) -> Result<Self, PngError>;
+    fn write_to_png<W: Write>(&self, stream: W) -> io::Result<()>;
+}
+
+impl ImageSurfacePng for ImageSurface {
+    fn create_from_png<R: Read>(stream: R) -> Result<ImageSurface, PngError> {
+        let mut env = ReadEnv{ reader: stream, error: None };
+        let surface: ImageSurface = unsafe { from_glib_full(ffi::cairo_image_surface_create_from_png_stream(
+            transmute(read_func::<R>), transmute(&mut env))) };
```
gkoz: It should not be necessary to transmute `read_func`, isn't its signature compatible with `cairo_read_func_t`?
Also in the interest of avoiding transmutes I'd prefer `&mut env as *mut ReadEnv as *mut c_void`.

src/image_surface_png.rs
vojtechkral: Hmm. I must've had some outdated nightly rust version accidentally... That's what I get for switching various rustc versions...

gkoz:
I'm still a little worried about error handling: `cairo_surface_write_to_png_stream` seems capable of returning non-zero results, including `NoMemory`.

GuillaumeGomez:
@gkoz: As long as they're wrapped, there should be no problem, no?

gkoz:
Unlike `create_from_png`, the present version of `write_to_png` ignores them.

GuillaumeGomez:
Right, I thought cairo handled it internally.

vojtechkral:
Well, I forgot to commit the file :-/
And yes the status should be checked too. I'll have a look at it tomorrow...

vojtechkral:
Ok, there it is I hope. Sorry for the delay. Squashing needed?

GuillaumeGomez:
Yes please.

gkoz:
Consider implementing `std::error::Error` for `Error` but I won't demand that.
Other than that :+1:

GuillaumeGomez:
@gkoz: Open an issue. :p

vojtechkral:
I'd rather not elaborate on the error handling, there seems to be a lot of space for design choices...

gkoz:
Fair enough.

Thanks for the PR!

PR #61, opened 2016-02-16
Bump cairo-sys-rs properly

gkoz:


gkoz:
Oops, looks like #62 needs to go in before this.

PR #66, opened 2016-02-28
Add lgpl-docs features and minor cleanups

gkoz:


GuillaumeGomez:
I can't merge it. Please update it.

GuillaumeGomez:
Also, can you squash the last commit?

gkoz:
It feels wrong to do that: they change separate entities; adjusting CI is not a part of adding those features.

GuillaumeGomez:
From my point of view, it's the same change. The first make the second mandatory.

gkoz:
Perhaps the CI commits should be phrased better. They certainly depend on the presence of the features but the point is quite different: preventing accidental checkins of embedded docs and secondarily detecting doctests breakage.

GuillaumeGomez:
I won't bother with you on this detail furthermore. I don't agree on your arguments but let's merge it. Great work! Thanks! :)

PR #67, opened 2016-03-01
Reexport IoError

gkoz:
It's as good as private right now, I'm surprised the compiler allows that.


GuillaumeGomez:
If it's not used from the outside, I'm not sure it should forbid it. However it should be public, so thanks! :)

PR #73, opened 2016-04-07
Bump versions

gkoz:


GuillaumeGomez:
You don't want to see if other issues come up before changing version?

gkoz:
I don't mean to release the crates, it's basically the same as https://github.com/gtk-rs/gtk/pull/229 to avoid the impression that master is the same version as on crates.

GuillaumeGomez:
Then all new builds will fail. How do you want to avoid that?

gkoz:
The last time we merged all bumps at the same time and then the builds became compatible once again.

GuillaumeGomez:
Then let's do it again! Thanks @gkoz!

PR #78, opened 2016-05-25
Make Surface struct fields public

kandeshvari:
Hi, 
I've create cairo-xcb bindings and it needed that Surface struct's fields to be a public. I create surface from a pointer and create context as:

```rust
    let csurface = xcb_surface_create(&conn, window, &visual, 150, 150);
    let surface = Surface(csurface);
    let cr = cairo::Context::new(&surface);
```

### DIFF
--- a/src/surface.rs
+++ b/src/surface.rs
@@ -13,7 +13,7 @@ use ffi::enums::{
 };
 
 #[derive(Debug)]
-pub struct Surface(*mut ffi::cairo_surface_t);
+pub struct Surface(pub *mut ffi::cairo_surface_t);
 
 impl Surface {
     pub fn status(&self) -> Status {


GuillaumeGomez:
I don't like setting fields public. Wouldn't it preferrable to create a `#[doc(hidden)]` function @gkoz?

gkoz:
You can import `glib::translate::FromGlibPtr` and use `Surface::from_glib_full(ptr)`.

kandeshvari:
@gkoz  In this case I should set the glib dependency in Cargo.toml. Maybe will better to add public constructor like `Surface::new(ptr)` to struct?

GuillaumeGomez:
I don't see the problem off adding a dependency (which you already have anyway).

kandeshvari:
What is the problem to make a field public or adding a constructor?

GuillaumeGomez:
Because we want it to remain opaque.

gkoz:
> What is the problem to make a field public

Basically, encapsulation.

> or adding a constructor?

The constructors are there, you just need to import a trait to access them. Since the consumers are not supposed to use them, extreme convenience of accessing them is not called for.

kandeshvari:
Maybe it's crazy but I've solved this problem my own way:

```rust
#[derive(Debug)]
pub struct SurfacePub(*mut cairo_surface_t);

let visual = find_visual(&conn, screen.root_visual()).unwrap();
let csurface = xcb_surface_create(&conn, window, &visual, 150, 150);
let surface = unsafe {&*((&SurfacePub(csurface) as *const SurfacePub) as *const Surface)};
let cr = cairo::Context::new(&surface);
```

gkoz:
I don't believe gratuitous unsafety (pretty sure you have a dangling pointer there) solves anything but I guess everyone has different priorities.

PR #81, opened 2016-06-28
Add xlib based create_surface functions to cairo-rs-sys and "xlib" feature flag

ludat:
I needed to use cairo xlib bindings and I though they could live here using a feature flag.

Not sure if the build.rs needs updating or the build.rs of the x11 crate solves any linking issue.

Hope it's useful.

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -6,7 +6,13 @@
 
 extern crate libc;
 
+#[cfg(feature = "xlib")]
+extern crate x11;
+
 use libc::{c_void, c_int, c_uint, c_char, c_uchar, c_double, c_ulong};
+
+#[cfg(feature = "xlib")]
+use x11::xlib;
 
 pub mod enums;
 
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Thanks for your PR!

First thing I wonder is: shouldn't you add a dependency to `libx11` in case the feature is activated?

gkoz:
Thanks for the PR. I'll add some formatting comments.

cairo-sys-rs/src/lib.rs
```diff
@@ -8,6 +8,12 @@ extern crate libc;
 
 use libc::{c_void, c_int, c_uint, c_char, c_uchar, c_double, c_ulong};
 
+#[cfg(feature = "xlib")]
+extern crate x11;
```
gkoz: By convention all `extern crate` should precede the `use` lines.

cairo-sys-rs/src/lib.rs
```diff
@@ -8,6 +8,12 @@ extern crate libc;
...
+
+#[cfg(feature = "xlib")]
+use x11::xlib::{Display, Drawable, Visual, Screen, Pixmap};
```
gkoz: This import could become a problem when more backends are added. I'd prefer importing just `xlib` and referring to these as e.g. `xlib::Display`.

cairo-sys-rs/src/lib.rs
```diff
@@ -462,4 +468,49 @@ extern "C" {
     pub fn cairo_image_surface_create_from_png_stream(read_func: cairo_read_func_t, closure: *mut c_void) -> *mut cairo_surface_t;
     #[cfg(feature = "png")]
     pub fn cairo_surface_write_to_png_stream(surface: *mut cairo_surface_t, write_func: cairo_write_func_t, closure: *mut c_void) -> Status;
+
+    #[cfg(feature = "xlib")]
+    pub fn cairo_xlib_surface_create(dpy: *mut Display, drawable: Drawable,
```
gkoz: If doing one argument per line, keep it consistent.

cairo-sys-rs/src/lib.rs
```diff
@@ -462,4 +468,49 @@ extern "C" {
...
+                                     visual: *mut Visual,
+                                     width: c_int,
+                                     height: c_int)
+     -> *mut cairo_surface_t;
```
gkoz: I think the prevailing style is to indent to arguments:
```
                                     height: c_int)
                                     -> *mut cairo_surface_t;
```

src/lib.rs
```diff
@@ -2,7 +2,7 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
-extern crate cairo_sys as ffi;
+pub extern crate cairo_sys as ffi;
```
gkoz: This bit seems to lack context. Generally the syntax is obsolete and it shouldn't be necessary.

ludat:
@gkoz fixed all issues, hope it's ok now.

@GuillaumeGomez the dependency is in the cargo.toml of cargo-sys because of cairo's abstractions the cairo crate can ignore x11. [little POC](https://github.com/ludat/rubar/blob/4f5350acadedc7d6717e583a98a3b94e646713f6/src/window.rs#L184)

That said, calling `cairo_xlib_create_surface` directly and handling pointers directly is not a great abstraction [(code)](https://github.com/ludat/rubar/blob/8e34391f28f17809aeedf263044d329d53309835/src/main.rs#L27) but this PR only adds things to cairo-sys-rs not to cairo-rs.

GuillaumeGomez:
You didn't understood what I asked but my question was badly written. Anyway, it doesn't really matter, I was just wondering about:

```
#[link(name = "libx11")] extern {}
```

But it seems unneeded so just ignore my question. :smiley_cat:

ludat:
@GuillaumeGomez yeah I misunderstood your question, the link x11 lives in the x11 crate which takes care of all the x11 bindings. I not 100% confident about this but it compiles and works.

GuillaumeGomez:
Ah right, didn't think about taking a look inside `x11` crate. Then yes, it's supposed to handle linking by itself.

gkoz:
Okay, looks good. Please squash and rebase on top of #82 (or wait for it to be merged).

In the bigger picture though I'm not sure how far further you can get: are there any high-level xlib bindings? We can't put functions which work with raw pointers in the cairo-rs crate.

ludat:
Well, AFAIK the cairo crate as a standalone crate can't be used without calling ffi functions directly and using pointers.

I'll wait until that gets merged and then squash/rebase

gkoz:
Unlike Xlib, there high-livel bindings for XCB. Do you specifically need to use Xlib?

ludat:
Not really, I just found more info about xlib+cairo than xcb+cairo, could you link the xcb lib?

gkoz:
I've edited the above comment, you might need to refresh the page.

ludat:
The issue is still the same, libraries don't expose pointers and cairo_XXX_surface_create wants a pointer. Maybe you could use GlibToPtr to introspect the container of xcb and extract the correct pointer or write a new trait MakeSurfaceFrom or something like that.

gkoz:
They do expose the pointers: http://rtbo.github.io/rust-xcb/xcb/base/struct.Connection.html#method.get_raw_conn, http://rtbo.github.io/rust-xcb/xcb/base/struct.StructPtr.html#structfield.ptr. Properly creating a `Connection` from a raw pointer doesn't seem possible though.

Cargo.toml
```diff
@@ -18,6 +18,7 @@ name = "cairo"
 [features]
 embed-lgpl-docs = ["gtk-rs-lgpl-docs"]
 png = ["cairo-sys-rs/png"]
+xlib = ["cairo-sys-rs/xlib"]
```
gkoz: I think we should drop this feature from here because there's no way for the crate to provide any Xlib support and you have to explicitly depend on the sys crate to use the functions added by this PR.

ludat: I completely agree.

ludat:
That's enough though, the raw pointer can be used for cairo_xcb_surface_create, I still think there should be a MakeSurface trait and change the type of Context::new` to `Context::new<T: MakeSurface>(T [...]`

Anyway I'll probably stick to the sys crates so I won't make any change out of that.

gkoz:
Merged #82 it should be fine to rebase and squash now.

ludat:
@gkoz this should be ready for merging now.

gkoz:
Thanks!

PR #82, opened 2016-06-29
Appveyor update

gkoz:


GuillaumeGomez:
I thought we already did it a while ago? Thanks a lot for fixing it!

Issue #88, opened 2016-11-01
Need to be able to Path::wrap() publically


federicomenaquintero:
I'm porting parts of librsvg to Rust; it uses Cairo internally.

At some point I get a cairo_path_t from the C code.  However, gtk-rs/cairo doesn't make ffi::cairo_path_t public.

I added a simple

  pub use ffi::cairo_path_t as cairo_path_t;

to cairo/lib.rs, so I can then

extern crate cairo;

extern "C" {
    fn rsvg_path_builder_copy_path (builder: *mut RsvgPathBuilder) -> *mut cairo::cairo_path_t;
}

fn my_stuff (builder: *mut RsvgPathBuilder) {
    let cairopath: *mut cairo::cairo_path_t;

    unsafe { cairopath = rsvg_path_builder_copy_path (builder); }
    let path = cairo::Path::wrap (cairopath);
    ... do stuff with path ...
}

at which point I have a proper cairo::Path out of its ::wrap().

I'm not sure if adding that "pub use ... as cairo_path_t" is the best solution, though it seemed easy enough :)  Ideas are appreciated.

federicomenaquintero:
By the way, I similarly envision the need to wrap a cairo_t when I get one from the C code.  Maybe this is a general pattern that needs to be made available - to wrap Cairo objects one gets from C, so that code can be migrated to Rust piece by piece?

federicomenaquintero:
You can look at the code here - https://github.com/GNOME/librsvg/blob/rustification/rust/src/marker.rs#L371

gkoz:
The `cairo_t` types are being imported from the [`cairo-sys`](https://github.com/gtk-rs/cairo/tree/master/cairo-sys-rs) crate and you can do the same in your code. Unless I'm missing some obvious problem with that.

BTW `Path::wrap` not being marked `unsafe` is a bug, looks like it hasn't been unified with the other types: [wrapping `cairo_t *` uses a trait](https://github.com/gtk-rs/cairo/blob/master/src/context.rs#L53-L66) defined in [`glib::translate`](http://gtk-rs.org/docs/glib/translate/index.html). With `Path` it would probably look like

```rust
let path = unsafe { from_glib_full(rsvg_path_builder_copy_path (builder)); }
```

[...]

PR #89, opened 2016-11-02
Refactor Path

gkoz:
Make `Path` a newtype over `cairo_path_t` and introduce `BoxedPath`, an owned pointer to `Path`.

`BoxedPath` has a library-specific destructor and should only be returned by this crate. Other crates can import `Path` and define their own `BoxedPath` variants with appropriate destructors.

This should solve the destructor mismatch pointed out in https://github.com/gtk-rs/cairo/issues/88#issuecomment-257510285.

@federicomenaquintero With this change you can import `Path` from this crate and define your own `BoxedPath` with a different destructor.

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -236,7 +236,7 @@ extern "C" {
     pub fn cairo_copy_path(cr: *mut cairo_t) -> *mut cairo_path_t;
     pub fn cairo_copy_path_flat(cr: *mut cairo_t) -> *mut cairo_path_t;
     pub fn cairo_path_destroy(path: *mut cairo_path_t);
-    pub fn cairo_append_path(cr: *mut cairo_t, path: *mut cairo_path_t);
+    pub fn cairo_append_path(cr: *mut cairo_t, path: *const cairo_path_t);
     pub fn cairo_has_current_point(cr: *mut cairo_t) -> cairo_bool_t;
     pub fn cairo_get_current_point(cr: *mut cairo_t, x: *mut c_double, y: *mut c_double);
     pub fn cairo_new_path(cr: *mut cairo_t);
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
All good for me.

gkoz:
Not handling version bumps today because that's likely to get messy...

federicomenaquintero:
Having FromGlibPtr and ToGlibPtr implemented for Path is very nice indeed - makes all the wrapping/unwrapping easier.

As for memory management...

Although I'm not very familiar with Rust idioms, I'm not very convinced about having BoxedPath with a hardcoded call to cairo_path_destroy().  If users of the library have their own sources of paths, wouldn't it be easier for them to have a way to box cairo_path_t with their own destructors?  Then gtk-rs/cairo could provide a default implementation with cairo_path_destroy() for the case where it really got the cairo_path_t from Cairo itself (or a library which presents a cairo_path_t obtained from Cairo itself).

Something like

let my_path = BoxedPath (my_cairo_path_t_pointer, my_destroy_fn);

The thing is, I don't need this kind of code in librsvg anymore :-P  I've changed it to handle path data by itself, without having cairo_path_data_t as an internal representation, and that also served to clean up a little bunch of code.  I don't know how common it is for apps or libraries to keep their own paths built by hand from cairo_path_data_t.  In any case, it seems easier to me to provide people with a way to box those with a specific destructor, instead of making people implement their own boxing.

EPashkin:
:+1:
@gkoz, Build fails on nightly due bug in cargo, it now fixed (https://github.com/gtk-rs/gtk/issues/404),
but you forgot replace last `Path::wrap` in `patterns.rs`, so GTK-3.18 fails too

sdroege:
What's with this one?

sdroege:
@GuillaumeGomez I guess this can be closed. No update since years and merge conflicts.

PR #90, opened 2016-11-04
Fixing typo in LineJoin enum

RazrFalcon:
There should be `Bevel` not `nBevel`. Reference: https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-line-join-t


GuillaumeGomez:
Thanks!

Whereas the tests fail, I'll still merge it since it doesn't seem to be related to your code. The last cargo version has changed the handling of Cargo.lock apparently...

GuillaumeGomez:
cc @gkoz

PR #92, opened 2016-11-09
add 2 png functions

ultimaRatio:
add basic read and write png functions (just had to give a filename).

### DIFF
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -78,6 +78,24 @@ impl ImageSurface {
     pub fn get_width(&self) -> i32 {
         unsafe { ffi::cairo_image_surface_get_width(self.to_glib_none().0) }
     }
+    
+    #[cfg(feature = "png")]
+    pub fn to_png(&self, filename : String) -> Status {
+       use std::ffi::CString;
+       match CString::new(filename) {
+           Ok(file_name) =>  unsafe { ffi::cairo_surface_write_to_png(self.to_glib_none().0, file_name.as_ptr()) },
+           _ => Status::WriteError
+       }
+    }
+    
+    #[cfg(feature = "png")]
+	pub fn from_png(filename : String) -> Option<ImageSurface> {
+       use std::ffi::CString;
+       match CString::new(filename.into_bytes()) {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Well, you just added FFI. It's useless as is.

ultimaRatio:
ok, sorry but i only use ffi.

ultimaRatio:
i've add the 2 small functions to the non-ffi code. 
You'll probably might want to change some thingss, as it may not be coherent with your code

src/image_surface.rs
```diff
@@ -78,6 +78,24 @@ impl ImageSurface {
     pub fn get_width(&self) -> i32 {
         unsafe { ffi::cairo_image_surface_get_width(self.to_glib_none().0) }
     }
+    
+    #[cfg(feature = "png")]
+    pub fn to_png(&self, filename : String) -> Status {
```
GuillaumeGomez: `filename` should be a `&str` or a `&Path` or even:

```rust
pub fn to_png<S>(&self, filename: S) -> Status {
where S: AsRef<Path>
```

src/image_surface.rs
```diff
@@ -78,6 +78,24 @@ impl ImageSurface {
...
+       use std::ffi::CString;
+       match CString::new(filename) {
+           Ok(file_name) =>  unsafe { ffi::cairo_surface_write_to_png(self.to_glib_none().0, file_name.as_ptr()) },
```
GuillaumeGomez: Since `file_name` isn't supposed to be a `String`, creating a `CString` is useless.

src/image_surface.rs
```diff
@@ -78,6 +78,24 @@ impl ImageSurface {
...
+           _ => Status::WriteError
+       }
+    }
+    
+    #[cfg(feature = "png")]
+	pub fn from_png(filename : String) -> Option<ImageSurface> {
```
GuillaumeGomez: All same as above.

src/image_surface.rs
```diff
@@ -78,6 +78,24 @@ impl ImageSurface {
...
+       use std::ffi::CString;
```
GuillaumeGomez: Indent broken.

EPashkin:
Can't check it now, but normally gtk file functions has different names for windows.
Its reason why no file read\write function in gtk (if I not wrong)

PR #93, opened 2016-11-27
Update crates.io version

GuillaumeGomez:


EPashkin:
Forgot cairo-sys-rs\Cargo.toml

GuillaumeGomez:
Ah indeed, good catch!

EPashkin:
:+1:
Thanks

PR #97, opened 2016-12-29
Remove CVec

daym:
CVec dev-depends on libc 0.1 which I can't compile using rustc 1.14.0 because of

```
   Compiling libc v0.1.12 (file:///tmp/guix-build-rust-libc-0.1.12.drv-0/libc-0.1.12)
error[E0554]: #[feature] may not be used on the stable release channel
  --> rust/src/liblibc/lib.rs:26:19
   |
26 | #![cfg_attr(test, feature(test))]
   |                   ^^^^^^^^^^^^^

error: aborting due to previous error
```

The current libc is 0.2 which is incompatible with 0.1.

Only 8 Rust crates in total seem to still use c_vec - and many moved away.

Therefore, I removed c_vec from cairo-rs.

Note that I'm a beginner to Rust. However, the compilation worked and all the tests run by `cairo test` still pass.

The parts that still used CVec were:

- src/context.rs: `copy_clip_rectangle_list` which calls an FFI function which returns newly allocated memory, eventually freed via impl Drop (by us). Therefore, the lifetime of the newly allocated memory is the same as ours.
- src/paths.rs: `iter`: the PathSegments are part of the Path and therefore live as long as the Path does.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -3,8 +3,8 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use glib::translate::*;
-use c_vec::CVec;
 use std::mem::transmute;
+use std::slice::from_raw_parts;
 use libc::{c_double, c_int};
 use ::paths::Path;
 use ::fonts::{TextExtents, TextCluster, FontExtents, ScaledFont, FontOptions, FontFace, Glyph};
@@ -26,12 +26,12 @@ use ffi::enums::{Status, Antialias, LineCap, LineJoin, FillRule};
 use ::patterns::{wrap_pattern, Pattern};
 use surface::Surface;
 
-pub struct RectangleVec {
+pub struct RectangleVec<'a> {
     ptr: *mut cairo_rectangle_list_t,
-    pub rectangles: CVec<Rectangle>,
+    pub rectangles: &'a[Rectangle], // Vec<Rectangle>,
 }
 
-impl Drop for RectangleVec {
+impl<'a> Drop for RectangleVec<'a> {
     fn drop(&mut self) {
         unsafe {
             ffi::cairo_rectangle_list_destroy(self.ptr);
@@ -355,7 +355,7 @@ impl Context {
 
             RectangleVec {
                 ptr: rectangle_list,
-                rectangles: CVec::new((*rectangle_list).rectangles,
+                rectangles: from_raw_parts((*rectangle_list).rectangles,
                                       (*rectangle_list).num_rectangles as usize),
             }
         }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,4 +36,3 @@ version = "0.1.1"
 
 [dependencies]
 libc = "0.2"
-c_vec = "~1.0"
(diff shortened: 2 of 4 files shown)


GuillaumeGomez:
Such a change is quite problematic due to the high unsafety of the from_raw_parts method. Personnally, I'm against this change. What do you think about it @EPashkin?

EPashkin:
@GuillaumeGomez, currently I also against merging this PR.
IMHO Better solution will be publishing `c_vec 1.0.13` which already updated libc dependence in code but just not published.

src/context.rs
```diff
@@ -26,12 +26,12 @@ use ffi::enums::{Status, Antialias, LineCap, LineJoin, FillRule};
 use ::patterns::{wrap_pattern, Pattern};
 use surface::Surface;
 
-pub struct RectangleVec {
+pub struct RectangleVec<'a> {
     ptr: *mut cairo_rectangle_list_t,
-    pub rectangles: CVec<Rectangle>,
+    pub rectangles: &'a[Rectangle], // Vec<Rectangle>,
 }
```
EPashkin: `RectangleVec` filled from `Context::copy_clip_rectangle_list` may outlive source, so use lifetime may cause problems.

daym: What do you mean? I thought that we freed RectangleVec ourselves.

Also, RectangleVec.rectangles is created at the same time as the RectangleVec instance and also freed at the same time as the RectangleVec instance. In short, RectangleVec and RectangleVec.rectangles have the same lifetime.

Am I wrong in this? Or do you mean that the RectangleVec instance can be copied? If so, that's bad...

EPashkin: Line `pub fn copy_clip_rectangle_list(&self) -> RectangleVec {` with your version of `RectangleVec` actually mean `pub fn copy_clip_rectangle_list<'a>(&'a self) -> RectangleVec<'a> {`
So borrow checker think that `RectangleVec` can't outlive `Context` object for that function was called.
In most cases is not problem, but if you need store this vector for long time borrow checker nags about lifetimes.
PS. No, `RectangleVec` instance can't be copied, as it don't realize `Clone, Copy` in cairo crate, and has private member, that disallow create objects of this type outside

daym: >So borrow checker think that RectangleVec can't outlive Context object for that function was called.

... what magic is this? But good to know. Yeah, my intent was to have it be a refcounted object independent of Context (I checked cairo's copy_clip_rectangle_list - it just mallocs a list). But apparently that works in a different way. Sorry.

EPashkin: https://doc.rust-lang.org/book/lifetimes.html#lifetime-elision
By way, @daym, it solves your problem if @GuillaumeGomez publish `c_vec` with upgraded `lib_c` dependence?

daym: It should. I'll test it.

GuillaumeGomez:
I'll try to do the update in the next hours.

GuillaumeGomez:
I merged the [PR](https://github.com/GuillaumeGomez/c_vec-rs/pull/16). Updating cairo dependency.

GuillaumeGomez:
@daym: Maybe you could try https://github.com/gtk-rs/cairo/pull/98 ?

PR #99, opened 2017-01-23
run rustfmt

evmar:
I want to do some more changes to this file, but I thought I'd first format it so my other changes don't affect whitespace.

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -16,27 +16,9 @@ use x11::xlib;
 
 pub mod enums;
 
-use enums::{
-    Status,
-    Antialias,
-    LineCap,
-    LineJoin,
-    FillRule,
-    FontSlant,
-    FontWeight,
-    TextClusterFlags,
-    FontType,
-    SubpixelOrder,
-    HintStyle,
-    HintMetrics,
-    Extend,
-    Filter,
-    PathDataType,
-    PatternType,
-    Format,
-    SurfaceType,
-    Operator,
-};
+use enums::{Status, Antialias, LineCap, LineJoin, FillRule, FontSlant, FontWeight, TextClusterFlags,
+            FontType, SubpixelOrder, HintStyle, HintMetrics, Extend, Filter, PathDataType,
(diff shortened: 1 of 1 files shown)


cairo-sys-rs/src/lib.rs
```diff
@@ -16,27 +16,9 @@ use x11::xlib;
...
-    HintMetrics,
-    Extend,
-    Filter,
-    PathDataType,
-    PatternType,
-    Format,
-    SurfaceType,
-    Operator,
-};
+use enums::{Status, Antialias, LineCap, LineJoin, FillRule, FontSlant, FontWeight, TextClusterFlags,
```
GuillaumeGomez: I'm against this change: makes git diff less readable.

GuillaumeGomez:
Well, I'm not sure this is very useful, but why not...

evmar:
I figured I'd send this one first to see how you felt about it.  I wasn't sure if you intentionally hadn't run rustfmt or not. 

It turns out that my other changes don't depend on this file being formatted, so I can try to undo this part of my other patch series (just sent as a pull request).  However, I'm not sure I can make my function layout the same as yours in that file because I don't know your style rules (which is the purpose of tools like rustfmt).

It's your project, so let me know if you want this change or not.  I understand either way.  :)

GuillaumeGomez:
I think most of the changes are good. There is just the one I pointed that I'm against. So if you remove it, I'll merge it/

evmar:
There's no point in using a tool to format your code if you have to hand
edit it afterwards. How about I drop this and then rebase my other change
without it?

GuillaumeGomez:
On your other branch:

```bash
git rebase -i HEAD~[number of commits]
```

Then remove the line where you have your commit and:

```bash
git push -f origin HEAD:your_branch
```

evmar:
OK, I will drop this.

For what it's worth, I think the eventual plan is for all rust code to be formatted with rustfmt.  For example, the rust compiler appears to use it for their code.  (Using it allows other nice things, such as automated transformations of source code by programs.)  So it'd be nice to think about whether it makes sense to format your library too.  But that is independent of my Windows change.

PR #100, opened 2017-01-23
implement wrappers for win32 surfaces

evmar:
These changes implement enough win32 surfaces for me to succeed at porting a small program of mine to Windows.

Some things I'm not sure about:
- do I bump the version number in cairo-sys-rs?
- I mostly just followed ImageSurface, is that the right thing to do?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,3 +37,6 @@ version = "0.1.1"
 [dependencies]
 libc = "0.2"
 c_vec = "~1.2"
+
+[target.'cfg(windows)'.dependencies]
+winapi = "0.2.7"
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -9,6 +9,9 @@ extern crate libc;
 #[cfg(feature = "xlib")]
 extern crate x11;
 
+#[cfg(windows)]
+extern crate winapi;
+
 use libc::{c_void, c_int, c_uint, c_char, c_uchar, c_double, c_ulong};
 
 #[cfg(feature = "xlib")]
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -96,3 +96,8 @@ mod patterns;
 mod rectangle;
 mod surface;
 mod matrices;
+
+#[cfg(windows)]
+mod win32_surface;
+#[cfg(windows)]
+pub use win32_surface::Win32Surface;
```
GuillaumeGomez: Missing backline.

src/win32_surface.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright 2017, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+#![cfg(windows)]
```
GuillaumeGomez: No need to put it here since you import this file only if we're building on windows.

cairo-sys-rs/src/lib.rs
```diff
@@ -9,34 +9,19 @@ extern crate libc;
...
-    HintMetrics,
-    Extend,
-    Filter,
-    PathDataType,
-    PatternType,
-    Format,
-    SurfaceType,
-    Operator,
-};
+use enums::{Status, Antialias, LineCap, LineJoin, FillRule, FontSlant, FontWeight, TextClusterFlags,
```
GuillaumeGomez: I'm still against this precise change.

GuillaumeGomez:
Seems globally good to me, thanks!

What about you @EPashkin?

Cargo.toml
```diff
@@ -37,3 +37,6 @@ version = "0.1.1"
 [dependencies]
 libc = "0.2"
 c_vec = "~1.2"
+
+[target.'cfg(windows)'.dependencies]
+winapi = "0.2.7"
```
EPashkin: Empty line at end, please.

EPashkin:
Look good

evmar:
OK, I rebased and cleaned up, take another look.

I also still am not sure if I should change version numbers.

Cargo.toml
evmar: Done, here and in the other file.

GuillaumeGomez:
@evmar: No you're not. Thanks a lot for your work!

PR #102, opened 2017-01-26
Add xcb functions

### LINKED ISSUE #101: XCB functions
Do you plan to add bindings to https://cairographics.org/manual/cairo-XCB-Surfaces.html, at least in cairo-sys? I need them to implement 29.2.5 at https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/xhtml/vkspec.html#VkXcbSurfaceCreateInfoKHR.

GuillaumeGomez:
Fixes #101.

Do you mind giving it a try @Yamakaky please?

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -55,6 +55,27 @@ pub struct cairo_antialias_t(c_void);
 pub struct cairo_line_join_t(c_void);
 #[repr(C)]
 pub struct cairo_line_cap_t(c_void);
+
+#[cfg(feature = "xcb")]
+#[repr(C)]
+pub struct cairo_device_t(c_void);
+#[cfg(feature = "xcb")]
+#[repr(C)]
+pub struct xcb_connection_t(c_void);
+#[cfg(feature = "xcb")]
+pub type xcb_drawable_t = u32;
+#[cfg(feature = "xcb")]
+pub type xcb_pixmap_t = u32;
+#[cfg(feature = "xcb")]
+#[repr(C)]
+pub struct xcb_visualtype_t(c_void);
+#[cfg(feature = "xcb")]
+#[repr(C)]
+pub struct xcb_screen_t(c_void);
+#[cfg(feature = "xcb")]
+#[repr(C)]
+pub struct xcb_render_pictforminfo_t(c_void);
+
 #[repr(C)]
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct cairo_rectangle_t {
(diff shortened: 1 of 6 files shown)


cairo-sys-rs/src/lib.rs
```diff
@@ -55,6 +55,27 @@ pub struct cairo_antialias_t(c_void);
 pub struct cairo_line_join_t(c_void);
 #[repr(C)]
 pub struct cairo_line_cap_t(c_void);
+
+#[cfg(feature = "xlib")]
```
Yamakaky: shouldn't it be xcb?

Yamakaky:
Did you forget to add `xcb_window_t`? Also, do you know how I can get a `xcb_visualtype_t` with only a `xcb_connection_t` and a `xcb_window_t`?
The corresponding Xlib code is here (please comment if I did something wrong): https://github.com/Yamakaky/cendre/blob/master/src/surface/xlib.rs#L28-L33

src/context.rs
```diff
@@ -88,25 +90,44 @@ impl Context {
         self.status().ensure_valid();
     }
 
+    /// Creates a new Context with all graphics state parameters set to default values
```
EPashkin: @GuillaumeGomez, Cairo docs not under GPL?

GuillaumeGomez: Oh damn, didn't pay attention that docs were regenerated!

GuillaumeGomez:
> Did you forget to add xcb_window_t? Also, do you know how I can get a xcb_visualtype_t with only a xcb_connection_t and a xcb_window_t?
The corresponding Xlib code is here (please comment if I did something wrong): https://github.com/Yamakaky/cendre/blob/master/src/surface/xlib.rs#L28-L33

I don't know either, that's why I made the inner types public. You can create them yourself from the xlib I suppose.

cairo-sys-rs/src/lib.rs
GuillaumeGomez: Yes indeed, thanks!

GuillaumeGomez:
So, removed the unwanted docs and added missing files. Sorry about that!

Yamakaky:
SGTM, otherwise.

Yamakaky:
When do you want to merge this?
Also there are still some functions missing, like Quartz. https://www.cairographics.org/manual/cairo-surfaces.html

GuillaumeGomez:
When you confirm it works for you I guess? We don't have much interest in it so you're the one to tell us in here. ;)

GuillaumeGomez:
Oh and could you open an issue for the other one as well please?

Yamakaky:
Since I couldn't find how to get an `xcb_visualtype_t` and XLib works, I didn't check. As it's only function signatures, it should be fine though.

GuillaumeGomez:
Fine then, let's merge!

PR #103, opened 2017-01-28
Implement missing methods for patterns and surfaces

federicomenaquintero:
These commits implement:
```
  Context.get_target() -> Surface
  Context.get_group_target() -> Surface
  Surface.create_similar() -> Surface
  SurfacePattern::create(surface)
  SurfacePattern.get_surface() -> Surface

```
I need these for librsvg :)

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -21,6 +21,7 @@ pub mod enums;
 
 use enums::{
     Status,
+    Content,
     Antialias,
     LineCap,
     LineJoin,
@@ -180,7 +181,7 @@ extern "C" {
     pub fn cairo_restore (cr: *mut cairo_t);
     pub fn cairo_get_target (cr: *mut cairo_t) -> *mut cairo_surface_t;
     pub fn cairo_push_group (cr: *mut cairo_t);
-    pub fn cairo_push_group_with_content (cr: *mut cairo_t, content: cairo_content_t);
+    pub fn cairo_push_group_with_content (cr: *mut cairo_t, content: Content);
     pub fn cairo_pop_group (cr: *mut cairo_t) -> *mut cairo_pattern_t;
     pub fn cairo_pop_group_to_source (cr: *mut cairo_t);
     pub fn cairo_get_group_target (cr: *mut cairo_t) -> *mut cairo_surface_t;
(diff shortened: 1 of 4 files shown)


cairo-sys-rs/src/lib.rs
```diff
@@ -21,6 +21,7 @@ pub mod enums;
 
 use enums::{
     Status,
+    Content,
```
EPashkin: As far I see `cairo_content_t` it just C enum, that has name to separate from other cairo enums.
Rust `Content` is equivalent, so IMHO better way just replace `cairo_content_t` with type alias `pub type cairo_content_t = enums::Content;` and remove this line, this way you don't need do full crate replace of `cairo_content_t`.
In gtk used different method: `KeyFileError` and `GKeyFileError` is completely separated but `KeyFileError` implement `ToGlib<GlibType = ffi::GKeyFileError>`, but `KeyFileError` resides in normal crate, not sys, so it not applied to this case.

federicomenaquintero: Oh, I put "Content" instead of "cairo_content_t" because that is the pattern used for Status/cairo_status_t already.

I see now that there is an empty

```
#[repr(C)]
pub struct cairo_content_t(c_void);

```

I'll remove that.

So would you prefer if I used "pub type cairo_content_t = enums::Content" and left all the function prototypes with cairo_content_t?

EPashkin: Sorry, I missed that you already do full replace (`cairo_content_t` just have few use).
I like to explicitly comment `cairo_content_t` with indicating replacement, but as it not done in other cases we can ignore here too and use PR as is.
@GuillaumeGomez, what do you think?

src/patterns.rs
```diff
@@ -14,12 +16,14 @@ use ffi::enums::{
 };
 use ffi;
 use ffi::{
-    cairo_pattern_t
+    cairo_pattern_t,
+    cairo_surface_t
```
GuillaumeGomez: Missing comma.

src/patterns.rs
```diff
@@ -14,12 +16,14 @@ use ffi::enums::{
...
 };
 use ::{
     Path,
     Matrix,
     MatrixTrait,
+    Surface
```
GuillaumeGomez: Missing comma.

cairo-sys-rs/src/lib.rs
GuillaumeGomez: @EPashkin: Except if you have strong feelings against this implementation, I'm ok with it.

src/patterns.rs
federicomenaquintero: Pushed the commas :)

GuillaumeGomez:
All good to me. Anything remaining for you @EPashkin?

EPashkin:
All ok.

GuillaumeGomez:
Then it's all good. Thanks @federicomenaquintero!

federicomenaquintero:
Thanks!  Also thanks for adding the missing exported Content/Extend/Filter; I think I already ran into that :)

PR #109, opened 2017-03-04
Update crates version

GuillaumeGomez:


EPashkin:
Don't see changes in sys toml

GuillaumeGomez:
Ah damn, I'll update my script.

EPashkin:
All other seems good

GuillaumeGomez:
Then I'll start the publications. Thanks @EPashkin!

EPashkin:
@GuillaumeGomez 
May be better just push your `GuillaumeGomez:crate` to `gtk-rs/cairo`.
As build fails always and I tracing your fork IMHO only need PR in gtk-rs about publishing new version.

GuillaumeGomez:
I prefer to keep both branches separated. It's not a problem if the build fails.

PR #110, opened 2017-03-09
Wrong feature name

Yamakaky:
It's a blocking bug for me, since the code doesn't compile without this fix.


GuillaumeGomez:
Oups indeed! Thanks for the fix!

GuillaumeGomez:
PS: Waiting for CIs anyway.

Yamakaky:
Could you do a release after that?

Yamakaky:
Seems good!

GuillaumeGomez:
Euh... For just one commit no, sorry.

Yamakaky:
Well the code doesn't compile, so I would do it.

GuillaumeGomez:
The release process is quite heavy. Once the "into" feature is merged though, I think we'll do another release. I think it'll be within a month but I can't guarantee it.

So until then, sorry for the disturbance and thanks again for your path!

PR #120, opened 2017-04-16
impl *GlibPtr* for font types, make raw pointer fields private

### LINKED ISSUE #119: Add unsafe from_raw method for FontOptions (and others)
I would like to suggest adding the following method to `FontOptions`:

```rust
    unsafe fn from_raw(ptr: *mut cairo_font_options_t) -> FontOptions {
        FontOptions(ptr)
    }
```

Side note: The inner pointer in `FontFace` is public. This is a bug, right? Since I can do unsafe things with this without using `unsafe` tag. Adding `from_raw` would be the correct alternative.

johncf:
I was unable to test this using pangocairo, because no matter [what I tried](https://github.com/RazrFalcon/pangocairo-rs/compare/master...johncf:cairo-fuk), I simply could not get the [`[replace]`](https://github.com/johncf/pangocairo-rs/blob/cairo-fuk/Cargo.toml#L33-L35) to work. It keeps showing `package replacement is not used` for both `cairo-rs` and `cairo-sys-rs`. May be I'm doing something wrong...

Fixes #119

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -622,7 +622,7 @@ impl Context {
 
     pub fn get_font_face(&self) -> FontFace {
         unsafe {
-            FontFace(ffi::cairo_get_font_face(self.0))
+            from_glib_none(ffi::cairo_get_font_face(self.0))
         }
     }
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -634,7 +634,7 @@ impl Context {
 
     pub fn get_scaled_font(&self) -> ScaledFont {
         unsafe {
-            ScaledFont(ffi::cairo_get_scaled_font(self.0))
+            from_glib_none(ffi::cairo_get_scaled_font(self.0))
         }
     }
 
--- a/src/fonts.rs
+++ b/src/fonts.rs
@@ -149,6 +149,24 @@ impl FontOptions {
     }
 }
 
+impl<'a> ToGlibPtr<'a, *const cairo_font_options_t> for &'a FontOptions {
+    type Storage = &'a FontOptions;
+
+    #[inline]
+    fn to_glib_none(&self) -> Stash<'a, *const cairo_font_options_t, &'a FontOptions> {
+        Stash(self.0, *self)
+    }
+}
+
+impl FromGlibPtrNone<*const cairo_font_options_t> for FontOptions {
+    #[inline]
+    unsafe fn from_glib_none(ptr: *const cairo_font_options_t) -> Self {
+        let tmp = ffi::cairo_font_options_copy(ptr);
+        assert!(!tmp.is_null());
+        FontOptions(tmp)
+    }
+}
+
 impl PartialEq for FontOptions {
     fn eq(&self, other: &FontOptions) -> bool {
         unsafe {
--- a/src/fonts.rs
+++ b/src/fonts.rs
@@ -161,9 +161,11 @@ impl<'a> ToGlibPtr<'a, *const cairo_font_options_t> for &'a FontOptions {
 impl FromGlibPtrNone<*const cairo_font_options_t> for FontOptions {
     #[inline]
     unsafe fn from_glib_none(ptr: *const cairo_font_options_t) -> Self {
-        let tmp = ffi::cairo_font_options_copy(ptr);
-        assert!(!tmp.is_null());
-        FontOptions(tmp)
+        let ptr = ffi::cairo_font_options_copy(ptr);
+        assert!(!ptr.is_null());
+        let tmp = FontOptions(ptr);
+        tmp.ensure_status();
+        tmp
     }
 }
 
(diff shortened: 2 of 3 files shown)


johncf:
Is there a way to automatically check for integration breakages with other gtk packages? (Since travis does not?)

EPashkin:
Please add FromPtrGlibFull
Pango on CI build examples so it checks that gtk-rs still compiles.
Although it is not works when 2 crates need changes and for external crates like pangocairo.
Your PR to pangocairo also better have change in call `pango_cairo_context_set_font_options` to use `options.to_glib_none().0`
As cargo don't support overrides for git dependencies to build it locally you need http://doc.crates.io/specifying-dependencies.html#overriding-with-local-dependencies but this need have all.

src/context.rs
```diff
@@ -622,7 +622,7 @@ impl Context {
 
     pub fn get_font_face(&self) -> FontFace {
         unsafe {
-            FontFace(ffi::cairo_get_font_face(self.0))
+            from_glib_none(ffi::cairo_get_font_face(self.0))
```
johncf: Btw, was this a bug? Since previously, `cairo_font_face_reference` was not called after calling `cairo_get_font_face`. So when `FontFace` gets dropped, `cairo_font_face_destroy` will get invoked possibly deallocating the original font_face while still in use.

Or am I doing [it](https://github.com/johncf/cairo/blob/d65b2531/src/fonts.rs#L269) wrong?

EPashkin: Yes it was bug, 
https://www.cairographics.org/manual/cairo-text.html#cairo-get-font-face cleary says that reference needed.

EPashkin:
PS. There typo in 3th commit name

EPashkin:
For pangocairo PR: better call `ensure_status` in `get_font_options`

EPashkin:
or better `cargo-sys.Status` need function `is_valid` and result checked with it and return None

src/context.rs
johncf: Looks like `ScaledFont` also has [the exact same bug](https://github.com/johncf/cairo/blob/f87f713c/src/context.rs#L637).

EPashkin: You right

src/fonts.rs
```diff
@@ -236,6 +262,32 @@ impl FontFace {
...
+        Stash(self.0, *self)
+    }
+}
+
+impl FromGlibPtrNone<*mut cairo_font_face_t> for FontFace {
+    #[inline]
+    unsafe fn from_glib_none(ptr: *mut cairo_font_face_t) -> Self {
+        let tmp = ffi::cairo_font_face_reference(ptr);
+        assert!(!tmp.is_null());
+        FontFace(tmp)
```
EPashkin: Seems this, ScaledFont and maybe all others need `ensure_status` check.

johncf: To prevent `CAIRO_STATUS_NO_MEMORY`?

EPashkin: Yes

EPashkin: And in from_glib_full too

EPashkin:
@johncf Thanks.

@GuillaumeGomez please, take look

johncf:
@EPashkin :+1: 

Any plans to replace those hidden `get_ptr` calls with `to_glib_none` calls? There are six definitions (and about 53 calls in total):

```
$ rg 'fn get_ptr'
src/paths.rs
19:    pub fn get_ptr(&self) -> *mut cairo_path_t {

src/patterns.rs
51:    fn get_ptr(&self) -> *mut cairo_pattern_t;
133:            fn get_ptr(&self) -> *mut cairo_pattern_t{

src/fonts.rs
79:    pub fn get_ptr(&self) -> *mut cairo_font_options_t {
210:    pub fn get_ptr(&self) -> *mut cairo_font_face_t {
311:    pub fn get_ptr(&self) -> *mut cairo_scaled_font_t {
```

Also, there are a few more public raw pointer fields:

```
$ rg '\(pub'
src/xcb.rs
29:pub struct XCBConnection(pub *mut ffi::xcb_connection_t);
67:pub struct XCBRenderPictFormInfo(pub *mut ffi::xcb_render_pictforminfo_t);
105:pub struct XCBScreen(pub *mut ffi::xcb_screen_t);
208:pub struct XCBVisualType(pub *mut ffi::xcb_visualtype_t);
248:pub struct Device(pub *mut ffi::cairo_device_t);
```

Leaving it public would mean changing it is "safe", but it's not. So these are incorrect APIs. It might be good to leave an issue open to fix this.

EPashkin:
Yes, it need be done but I don't have active plans for fixing it (not sure about @GuillaumeGomez).
As for XCB funcs it was added by @Yamakaky request in https://github.com/gtk-rs/cairo/pull/102 not sure how he use it.

EPashkin:
@GuillaumeGomez please, take look

GuillaumeGomez:
Good for me, so if it's good for @EPashkin as well, I'll merge.

PR #122, opened 2017-04-22
Proper glib bindings for font types

johncf:


johncf:
`FontOptions` commit would need some careful reviewing. I changed the `ffi` signatures based on documentation found [`FontOptions`](https://developer.gnome.org/cairo/stable/cairo-cairo-font-options-t.html), [`cairo_set_font_options`](https://developer.gnome.org/cairo/stable/cairo-text.html#cairo-set-font-options) and [`cairo_scaled_font_create`](https://developer.gnome.org/cairo/stable/cairo-cairo-scaled-font-t.html#cairo-scaled-font-create).

src/font/font_options.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn new() -> FontOptions {
+        let mut font_options: FontOptions = unsafe {
+            from_glib_full(ffi::cairo_font_options_create())
+        };
+        font_options.ensure_status();
+        font_options
+    }
+
+    #[doc(hidden)]
+    pub fn get_ptr(&self) -> *const cairo_font_options_t {
```
EPashkin: IMHO this and `get_ptr_mut` need `deprecated` attribute and must not used in `cairo`

src/font/font_options.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    #[doc(hidden)]
+    pub fn get_ptr_mut(&mut self) -> *mut cairo_font_options_t {
+        self.to_glib_none_mut().0
+    }
+
+    pub fn ensure_status(&mut self) {
+        let status = unsafe {
+            ffi::cairo_font_options_status(self.get_ptr_mut())
```
EPashkin: This don't need mutable pointer.
If  `mut_override(self.to_glib_none().0)` used then even don't need change ffi part.

src/font/font_options.rs
```diff
@@ -0,0 +1,115 @@
...
+        };
+        status.ensure_valid()
+    }
+
+    pub fn merge(&mut self, other: &mut FontOptions) {
+        unsafe {
+            ffi::cairo_font_options_merge(self.get_ptr_mut(), other.get_ptr_mut())
```
EPashkin: IMHO `other` don't need be mutable

src/font/font_options.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    pub fn hash(&self) -> u64{
+        unsafe {
+            ffi::cairo_font_options_hash(self.get_ptr()) as u64
+        }
+    }
+
+    pub fn set_antialias(&mut self, antialias: Antialias) {
+        unsafe {
+            ffi::cairo_font_options_set_antialias(self.get_ptr_mut(), antialias)
```
EPashkin: Not sure this need `&mut self`.

src/font/font_options.rs
johncf: So, replace their occurrences with `to_glib_none*` calls?

src/font/font_options.rs
johncf: Since `FontOptions` is a `Boxed` object, and not `Shared`, I thought this might be the proper way to do it. And gnome docs has meticulously marked `get_*` pointers as `const` for `font_options` unlike other objects that I found.

src/font/font_options.rs
```diff
@@ -0,0 +1,115 @@
...
+    HintStyle,
+    HintMetrics,
+};
+use ffi::cairo_font_options_t;
+
+glib_wrapper! {
+    pub struct FontOptions(Boxed<ffi::cairo_font_options_t>);
+
+    match fn {
+        copy => |ptr| ffi::cairo_font_options_copy(ptr),
```
EPashkin: Maybe this need contains checking status

src/font/font_options.rs
EPashkin: Yes

src/font/font_options.rs
johncf: For the pointer returned by `ffi::cairo_font_options_copy(ptr)`?

src/font/font_options.rs
EPashkin: From rust side this function don't change pointer.
And `to_glib_none_mut` used mainly for that.
@GuillaumeGomez what you think?

GuillaumeGomez: I'd prefer to keep the `&mut self` in here.

src/context.rs
```diff
@@ -600,23 +600,23 @@ impl Context {
...
+            ffi::cairo_set_font_options(self.0, options.to_glib_none().0)
         }
     }
 
     pub fn get_font_options(&self) -> FontOptions {
-        let out = FontOptions::new();
+        let mut out = FontOptions::new();
         unsafe {
-            ffi::cairo_get_font_options(self.0, out.get_ptr());
+            ffi::cairo_get_font_options(self.0, out.to_glib_none_mut().0);
```
EPashkin: IMHO this also don't changes internal pointer

src/font/font_options.rs
EPashkin: Yes

src/context.rs
EPashkin: But if @GuillaumeGomez prefer left &mut for setter this don't need change

johncf: It doesn't change the internal pointer, but it does modify the underlying data structure, right? So giving it a `*mut` might be the "better written" (or "better specified") thing to do?

PS: I'm not entirely sure when `*const` vs `*mut` should be used.

EPashkin:
@johncf Big thanks for this.
IMHO only last is to confirm that this crate still working (FontOption clone, etc. ). Can you do it?

src/context.rs
GuillaumeGomez: Yes, my point in here is trying to force user to have a mutable access if the underlying data is modifier (unlike get methods/functions).

EPashkin:
Forgot say about removed `get_ptr/get_ptr_mut`.
I proposed leave it but add `deprecated` attribute to it.
But as this crate still not `1.0.0` this not big problem for me.
@GuillaumeGomez It need be restored?

johncf:
> only last is to confirm that this crate still working (FontOption clone, etc. ).

By writing automatic test cases to check whether they work? I'm not very sure how to check their validity, though. Are there any test cases already in the repo that I can use as a template?

GuillaumeGomez:
> I proposed leave it but add `deprecated` attribute to it.

This is actually an excellent idea!

> But as this crate still not 1.0.0 this not big problem for me.

Well, it is for me. ;) If we have "stable" code enough, it'll only be easier for us to finally release a `1.0` version.

> @GuillaumeGomez It need be restored?

What do you mean?

> By writing automatic test cases to check whether they work? I'm not very sure how to check their validity, though. Are there any test cases already in the repo that I can use as a template?

I think the "easiest" way to confirm this is adding an example to the `examples` repo. Or maybe oyu had something else in mind @EPashkin?

johncf:
> I proposed leave it but add deprecated attribute to it.

Well this is a breaking change anyway. So, if you're following semantic versioning, next release after this is merged should be `0.2.0` (you're welcome :smile: ). So I don't think deprecating an already undocumented function would make sense.

johncf:
> I think the "easiest" way to confirm this is adding an example to the examples repo.

This would take me some time. I wouldn't be confident "using" this API yet. The only examples I have created using gtk-rs were by referring to examples in C code. If I could find any (or if you could point me to one), I'll add them.

GuillaumeGomez:
Never used it myself. Looks like it'll be fun haha.

EPashkin:
As I don't use cairo at all, so for me is good even if @johncf say that it works :wink:
Updated `example\cairotest.rs` is better

GuillaumeGomez:
@EPashkin: It's good for me as is, do I merge?

EPashkin:
@GuillaumeGomez  Yes if you agree with completely removed `get_ptr/get_ptr_mut`

johncf:
> As I don't use cairo at all, so for me is good even if @johncf say that it works :wink:

It works!! :crossed_fingers: 

Well, one simple example I could think of is to `get_font_options` from a cairo context, change antialias or something, and `set_font_options`. May be do a clone too while at it!! `ScaledFont` API looks scary! I'll see what I can do after a few days.

GuillaumeGomez:
@EPashkin: I'm just a bit worried about performance issues. If you don't think it'll be too much, then let's merge.

EPashkin:
👍 for merge.

johncf:
@GuillaumeGomez Could you specify the change you suspect to have introduced a performance regression.

GuillaumeGomez:
@johncf: It's quite small, but repeated, it *could* be big. `get_ptr*` was shorter than `to_glib_*` functions. And GUI performance is "simple" to check: as long as it's fluid, it's ok. :)

Thanks to both of you!

johncf:
> `get_ptr*` was shorter than `to_glib_*` functions.

Those will most likely get optimized out by LLVM gods!

> And GUI performance is "simple" to check: as long as it's fluid, it's ok. :)

No, seriously, in an automated test-like manner. Is there some way to attach a number to the current performance? For instance, it would be FPS for games.

GuillaumeGomez:
For GTK it's very complicated. I had a few discussions about this recently and it seems there isn't any tool "usable" enough for that unfortunately. :-/

johncf:
The closest thing I could find was [LDTP](https://en.wikipedia.org/wiki/Linux_Desktop_Testing_Project) from this [mailing list](https://mail.gnome.org/archives/gtk-devel-list/2005-June/msg00139.html). Just for testing, not benchmarking.

PR #126, opened 2017-05-01
Implement ToGlibPtrMut for Context

selaux:
Not totally sure if this is correct, because I feel like I've no idea what I'm doing 😉 . I'm currently in the process of generating bindings for rsvg and this Trait seems to be necessary to be able to pass a cairo context to rsvg.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -50,6 +50,16 @@ impl<'a> ToGlibPtr<'a, *mut ffi::cairo_t> for &'a Context {
     }
 }
 
+impl<'a> ToGlibPtrMut<'a, *mut ffi::cairo_t> for Context {
+    type Storage = &'a mut Self;
+
+    #[inline]
+    fn to_glib_none_mut(&'a mut self) -> StashMut<'a, *mut ffi::cairo_t, Self> {
+        let ptr: *mut Context = &mut *self;
+        StashMut(ptr as *mut ffi::cairo_t, self)
+    }
+}
+
 impl FromGlibPtrNone<*mut ffi::cairo_t> for Context {
     #[inline]
     unsafe fn from_glib_none(ptr: *mut ffi::cairo_t) -> Context {


EPashkin:
@selaux, IMHO this trait unneeded.
If you need manual call some function that accepts `*mut cairo_t`, you can pass it with `mut_override(context.from_glib_none().0)`
If problem in generated code, you can mark parameter [`const`](https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L400) and Gir do same for you

selaux:
That seems to work as well... 👍 

BTW is there any concise documentation on how to make your own bindings and what to take into account to do that? Because currently it kind of feels like tapping in the dark. E.g. I have unimplemented functions like follows and no idea how to automatically bind them or create bindings manually:

```
//pub fn new_from_data(data: /*Unimplemented*/&CArray TypeId { ns_id: 0, id: 3 }, data_len: /*Unimplemented*/Fundamental: Size) -> Result<Handle, Error> {
    //    unsafe { TODO: call ffi::rsvg_handle_new_from_data() }
    //}
```

EPashkin:
There some http://gtk-rs.org/docs-src/tutorial but it not cover creating own crates.
For manual you can look to sources for examples or/and implement missing functions.
For `rsvg_handle_new_from_data` function look something like this
```
pub fn new_from_data(data: &[u8]) -> Result<Handle, Error> {
assert_initialized_main_thread!();
unsafe {
let mut error = ptr::null_mut();
let handle = ffi::rsvg_handle_new_from_data(data.as_ptr() as *mut _, data.len() as _, &mut error);
if error.is_null() { Ok(from_glib_full(handle)) } else { Err(from_glib_full(error)) }
}
}
```

GuillaumeGomez:
@EPashkin: Do you think it might be worth adding a part explaining how to bind a library based on gtk-rs' crates?

EPashkin:
You can ignore `assert_initialized_main_thread`.
It used for static members to be sure that it called from thread where `gtk::init` or `gdk::init` called.
Also you can copy this macro from https://github.com/gtk-rs/gtk/blob/master/src/rt.rs#L21-L32 and use function from gtk::rt or gdk:rt (depend what initialization needed to work with rsvg)

EPashkin:
@GuillaumeGomez This doc is needed but IMHO it not belong tutorial as it need cover too many problem, not sure that we ready to write it now.

GuillaumeGomez:
I don't either, but we should try to do it some day in the future. :)

PR #142, opened 2017-07-15
Update for non-generic pointer array impls

sdroege:
In addition it might make sense to also implement the new traits for the
types that don't use glib_wrapper!()

See https://github.com/gtk-rs/glib/pull/198

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,9 +6,12 @@ extern crate cairo_sys as ffi;
 extern crate libc;
 extern crate c_vec;
 
-#[cfg(feature = "glib")]
+#[cfg(feature = "use_glib")]
 #[macro_use]
 extern crate glib;
+
+#[cfg(feature = "use_glib")]
+extern crate glib_sys as glib_ffi;
 
 pub use ffi::enums;
 pub use ffi::cairo_rectangle_t as Rectangle;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,8 @@ png = ["cairo-sys-rs/png"]
 xcb = ["cairo-sys-rs/xcb"]
 purge-lgpl-docs = ["gtk-rs-lgpl-docs"]
 v1_12 = ["cairo-sys-rs/v1_12"]
-default = ["glib"]
+use_glib = ["glib", "glib-sys"]
+default = ["use_glib"]
 
 [build-dependencies.gtk-rs-lgpl-docs]
 git = "https://github.com/gtk-rs/lgpl-docs"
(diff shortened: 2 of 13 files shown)


src/lib.rs
```diff
@@ -6,10 +6,14 @@ extern crate cairo_sys as ffi;
...
 extern crate c_vec;
 
-#[cfg(feature = "glib")]
+#[cfg(feature = "use_glib")]
 #[macro_use]
 extern crate glib;
 
+#[cfg(feature = "use_glib")]
+#[macro_use]
+extern crate glib_sys as glib_ffi;
```
EPashkin: Hm, glib_sys don't have macros

EPashkin:
Agree that implementing traits without glib will be good but currently IMHO better postpone this until all gtk-rs crates will be updated (if this don't make cairo build fails without glib).

sdroege:
> Agree that implementing traits without glib will be good but currently IMHO better postpone this until all gtk-rs crates will be updated (if this don't make cairo build fails without glib).

I meant implementing the traits that are not implemented generically anymore, and would be implemented by glib_wrapper!(). But here there are types that don't use glib_wrapper!() and currently implement ToGlibPtr and others manually.

EPashkin:
Ah, I missed point, yes it will be needed if it used as array.
Check gtk, gdk show that it not case for cairo and pango.

sdroege:
I think it's fine nowadays, but will know once I got everything compiling (currently at gdk, then gtk. gdk has some problems)

src/lib.rs
EPashkin: @sdroege, you forgot fix this.

EPashkin:
And some use need `#[cfg(feature = "use_glib")]`
```
warning: unused import: `std::ptr`
 --> src\font\font_options.rs:5:5
  |
5 | use std::ptr;
  |     ^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

warning: unused import: `std::mem`
 --> src\font\font_options.rs:6:5
  |
6 | use std::mem;
  |     ^^^^^^^^

warning: unused import: `std::ptr`
 --> src\font\font_face.rs:5:5
  |
5 | use std::ptr;
  |     ^^^^^^^^

warning: unused import: `std::mem`
 --> src\font\font_face.rs:6:5
  |
6 | use std::mem;
  |     ^^^^^^^^

warning: unused import: `std::mem`
 --> src\font\scaled_font.rs:6:5
  |
6 | use std::mem;
  |     ^^^^^^^^
```

sdroege:
Done, thanks

EPashkin:
In travis log even more warnings.

sdroege:
No warnings here in that configuration: it requires the glib pull request for the warnings to disappear.

EPashkin:
Forgot to say that warnings appeared when build without glib

sdroege:
I don't get any warnings here when build with --disable-default-features

EPashkin:
Sorry, forgot that it builds on old glb. All good

PR #146, opened 2017-08-09
ImageSurface: return a Result<ImageSurface, Status> from the creation functions

federicomenaquintero:
When Cairo gets asked to create a surface, it will always return a
non-null cairo_surface_t*, but its cairo_surface_status() may not be
CAIRO_STATUS_SUCCESS.

For the Rust-side surface creation functions, we now return a
Result<ImageSurface, Status>.  If Ok(), then the surface is in
Status::Success.  If Err(), we return the status from Cairo's error
surface.

This means that "image_surface = from_glib_full(ptr)" needs to be
passed a surface that is *not* in error.  This function will panic if
passed a Cairo surface that is in an error state.

### DIFF
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -30,25 +30,32 @@ impl ImageSurface {
         }
     }
 
-    pub unsafe fn from_raw_full(ptr: *mut ffi::cairo_surface_t) -> ImageSurface {
-        Self::from(Surface::from_raw_full(ptr)).unwrap()
+    pub unsafe fn from_raw_full(ptr: *mut ffi::cairo_surface_t) -> Result<ImageSurface, Status> {
+        let surface = Self::from(Surface::from_raw_full(ptr)).unwrap();
+        let status = surface.status();
+        match status {
+            Status::Success => Ok(surface),
+            _ => Err(status)
+        }
     }
 
-    pub fn create(format: Format, width: i32, height: i32) -> ImageSurface {
+    pub fn create(format: Format, width: i32, height: i32) -> Result<ImageSurface, Status> {
         unsafe { Self::from_raw_full(ffi::cairo_image_surface_create(format, width, height)) }
     }
 
     pub fn create_for_data<F>(data: Box<[u8]>, free: F, format: Format, width: i32, height: i32,
-                              stride: i32) -> ImageSurface
+                              stride: i32) -> Result<ImageSurface, Status>
     where F: FnOnce(Box<[u8]>) + 'static {
         assert!(data.len() >= (height * stride) as usize);
         unsafe {
             let mut data = Box::new(AsyncBorrow::new(data, free));
             let ptr = (*data).as_mut().as_mut_ptr();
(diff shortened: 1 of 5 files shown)


EPashkin:
`image_surface_png.rs` and maybe `win32_surface.rs` also need changes.

federicomenaquintero:
> image_surface_png.rs and maybe win32_surface.rs also need changes.

Oops, of course.  Sorry, I forgot to cargo test with --features png.  I'll add those changes.

federicomenaquintero:
I've pushed the changes to image_surface_png.rs and win32_surface.rs.  A couple of notes:

# PNG

There are new tests and they pass.  However, for an invalid PNG I'm getting different errors from Cairo depending on what byte on the PNG data I make invalid:  sometimes I get NoMemory, sometimes ReadError... I thought those would always give PngError?  Maybe this is a bug in Cairo or libpng?

# Win32Surface

I can't test the changes to win32_surface.rs, unfortunately.  I'm not sure that e.g. Win32Surface::create() is correct, as cairo_win32_surface_create() can return NULL as well as a Cairo error surface in case of errors.  Unfortunately I don't know the correct semantics there to give the correct error type in the Rust API.

src/image_surface.rs
```diff
@@ -30,25 +30,32 @@ impl ImageSurface {
...
         unsafe {
             let mut data = Box::new(AsyncBorrow::new(data, free));
             let ptr = (*data).as_mut().as_mut_ptr();
-            let surface = ImageSurface::from_raw_full(
+            let r = ImageSurface::from_raw_full(
                 ffi::cairo_image_surface_create_for_data(ptr, format, width, height, stride));
-            surface.set_user_data(&IMAGE_SURFACE_DATA, data).unwrap();
-            surface
+            match r {
+                Ok(surface) => surface.set_user_data(&IMAGE_SURFACE_DATA, data).map (|_| surface),
```
GuillaumeGomez: Extra whitespace `.map (`

GuillaumeGomez:
> I thought those would always give PngError? Maybe this is a bug in Cairo or libpng?

I prefer the dark magic theory.

Otherwise the current windows function seems normal/correct to me. Appveyor will confirm it.

EPashkin:
Build failing due examples don't ready for API change.
Thanks, @federicomenaquintero 
👍 for merge.

GuillaumeGomez:
@EPashkin: You want to update the examples or should I?

EPashkin:
I do it.

GuillaumeGomez:
Ok, thanks!

federicomenaquintero:
Thanks for merging and updating the examples!

I found the reason for the NoMemory error in case of a corrupted PNG.  It's a bug in Cairo (not that libpng makes its life particularly easy).  I've just filed https://bugs.freedesktop.org/show_bug.cgi?id=102142 about this.

GuillaumeGomez:
"How to improve libraries by binding them." by @federicomenaquintero ;)

Issue #145, opened 2017-08-09
Need a wrapper for cairo_format_stride_for_width()


federicomenaquintero:
`ImageSurface.create_for_data()` calls `cairo_image_surface_create_for_data()`, and its `stride` parameter:

1. Needs to be aligned to 32 bits

1. Must have an absolute value greater than or equal to `cairo_format_stride_for_width()` for the corresponding image (negative strides are allowed for bottom-to-top images).

The only place I could find that actually figures out a `stride` is in `examples/src/cairo-threads.rs`, but that gets it from an existing `image_surface.get_stride()` - it doesn't create a buffer on its own.

So, for people who actually want to allocate their own buffers to pass to `ImageSurface.create_for_data()`, I think we need a wrapper for `cairo_format_stride_for_width()`.

GuillaumeGomez:
It'd make things simpler indeed.

federicomenaquintero:
This can be closed now that it's merged.

PR #147, opened 2017-08-11
Add Format::stride_for_width() as a wrapper for cairo_format_stride_f…

federicomenaquintero:
…or_width()

https://github.com/gtk-rs/cairo/issues/145

### DIFF
--- a/cairo-sys-rs/src/enums.rs
+++ b/cairo-sys-rs/src/enums.rs
@@ -4,6 +4,8 @@
 
 use std::fmt::{Error, Debug};
 use std::ffi::CStr;
+use std::i32;
+use std::u32;
 
 #[repr(C)]
 #[derive(Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


federicomenaquintero:
Just so I don't delete them accidentally, these are the doc comments:

```
    /// Computes the distance between rows, in bytes, that an image
    /// of the specified `width` would have for the `Format`.
    ///
    /// Cairo requires that each row in image data be aligned in a
    /// particular way for each format.  You can pass the number
    /// inside the `Result` for this function to
    /// `ImageSurface::create_with_data()` as the `stride` argument.
    ///
    /// # Errors
    /// If the supplied `width` is too large for Cairo, this function
    /// will return `Err(())`.
    ///
    /// # Panics
    /// This function will panic if `width` is greater than i32::MAX.
    /// This is a limitation in Cairo.
```

EPashkin:
👍 Thanks

GuillaumeGomez:
Thanks!

PR #153, opened 2017-09-07
Mark `ImageSurface` and `PDFSurface` as `Send`

fengalin:
While following [this guide](https://www.cairographics.org/threaded_animation_with_cairo/), I found that `Surface` wasn't `Send`. I assume that if it is possible to share it between threads in C, it should also be possible in Rust.

### DIFF
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -150,6 +150,8 @@ impl Clone for ImageSurface {
     }
 }
 
+unsafe impl Send for ImageSurface {}
+
 pub struct ImageSurfaceData<'a> {
     surface: &'a mut ImageSurface,
     slice: &'a mut [u8],
--- a/src/pdf_surface.rs
+++ b/src/pdf_surface.rs
@@ -68,6 +68,8 @@ impl Clone for PDFSurface {
     }
 }
 
+unsafe impl Send for PDFSurface {}
+
 #[cfg(feature = "glib")]
 impl<'a> ToGlibPtr<'a, *mut ffi::cairo_surface_t> for PDFSurface {
     type Storage = &'a Surface;


fengalin:
Question: shouldn't `cairo-sys-rs` be part of `gtk-rs/sys`? `gtk-rs/sys/gdk-sys` and `gtk-rs/sys/gtk-sys`, depend on it. So when I modified `cairo`, I had to modify `gtk-rs/sys` to use my repository and ended up editing dependencies in: `gio`, `glib`, `gtk`, `gdk`, `gdk-pixbuf`, `pango` (and `gstreamer` and `gstreamer-sys` since my project also depends on `gstreamer`).

GuillaumeGomez:
> Question: shouldn't `cairo-sys-rs` be part of `gtk-rs/sys`?

No, it doesn't have a gir file so it's on its own.

fengalin:
All right! What about a separate crate then? ;) IMHO, it feels strange to modify `glib` and all the crates that depend on it for a simple modification in `cairo-rs`.

Anyway, that was just a mere comment. What do you think about the PR?

fengalin:
@GuillaumeGomez what do you think about this PR?

GuillaumeGomez:
For me it seems fine but I'd prefer to have the confirmation from @EPashkin and @sdroege. :)

sdroege:
I don't think this is safe, and the Cairo documentation there is at least misleading and outdated. See also https://developer.gnome.org/gdk3/stable/gdk3-Threads.html#gdk-threads-enter , those functions are deprecated since years.

While it is safe to send Cairo surfaces backed by a threadsafe underlying surface to another thread, this is not generally the case. With X11 you can in theory make it work (but it's not supported by GDK/GTK anymore, see above). The image surface is also always fine, being backed by just some memory. With Windows you can completely forget it and can't make it work (and the code on the Cairo website never worked on Windows).

sdroege:
tl;dr: This PR is wrong, but you can do something like this for image surfaces. And don't use GTK/GDK from multiple threads, ever :)

fengalin:
I don't use the deprecated functions you mentioned, the guide was just a source of inspiration. I'm implementing a double buffer mechanism, which by design must be multithreaded. Everything GTK related is done in the GTK thread. The double buffer mechanism is shared between threads, one doing the rendering using Cairo, the other being the GTK thread which handles drawing to a drawing area.

If this PR is wrong, how am I supposed to send my Cairo surface between the 2 threads?

fengalin:
Or are you saying that I should find a different design to work this around because this modification could be unsafe in an uncontrolled environment?

sdroege:
What kind of surface are you using for your double buffering implementation?

fengalin:
I'm using `cairo::ImageSurface`. I think I finally got your first comment. Can I safely set the `Send` trait on `cairo::ImageSurface` instead of the general `cairo::Surface`?

sdroege:
Yes, also the PNG, PDF, SVG surfaces should be safe to mark Send (but not Sync).

fengalin:
Ok! Thanks

fengalin:
Update since last commit:
- `Send` removed from `Surface`
- `Send` added to `ImageSurface` and `PDFSurface`

`PNGSurface` isn't a dedicated type, there is a convenience function `PNGSurface::create_from_png` that builds an `ImageSurface` using `cairo_image_surface_create_from_png_stream`.

`SVGSurface` is not available in **gtk-rs/cairo** yet.

EPashkin:
👍 for this now.

GuillaumeGomez:
@sdroege: Good for you as well?

sdroege:
Sure, that's exactly what I wrote before :)

GuillaumeGomez:
Then let's merge. Thanks!

PR #154, opened 2017-09-08
Add T: io::Write support to PDFSurface

meh:
When generating PDFs having to go around a temporary file is kind of annoying, I couldn't come up with a nice way to just expose the closure way but at least you can use `io::Write` to write to a `Vec<u8>` and similar.

I'm not entirely sure how safe the approach I took because I don't know how the whole GLib/`Clone` business is, but I think it *should* be safe.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ name = "cairo"
 [features]
 purge-lgpl-docs = ["gtk-rs-lgpl-docs"]
 png = ["cairo-sys-rs/png"]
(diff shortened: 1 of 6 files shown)


meh:
Actually, this wouldn't work for `io::Cursor`s wrapping `&mut [u8]` or `&mut Vec<u8>` because of the `'static` bound, but it would also be a bad idea to add a lifetime parameter to `PDFSurface`, maybe a new type for it?

meh:
Superseded by #165.

PR #156, opened 2017-10-07
Dox

GuillaumeGomez:


GuillaumeGomez:
cc @EPashkin

EPashkin:
I thought that possible problem with `dox` dependence on `glib` and it may work wrong without "use_glib".
But now I think that it not problem actually.
So 👍

GuillaumeGomez:
Yeay!

PR #158, opened 2017-10-21
Fix dox feature

GuillaumeGomez:
cc @EPashkin

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -11,6 +11,13 @@ extern crate x11;
 
 #[cfg(windows)]
 extern crate winapi;
+#[cfg(all(not(windows), feature = "dox"))]
+pub mod winapi {
+   use libc::c_void;
+
+   #[repr(C)]
+   pub struct HDC(c_void);
+}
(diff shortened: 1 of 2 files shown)


EPashkin:
Yes, it some solution, but docs will link `Display` or others to local version :(
And `surface.rs` also need add two `feature = "dox"` to remove errors.
IMHO better wait for normal fix, maybe fixed x11 will published today.

GuillaumeGomez:
Let's wait then.

GuillaumeGomez:
Ok, updated this one. Thanks for the fix in the x11 crate! :)

GuillaumeGomez:
@EPashkin: Do you have a 32 bits windows by any chance? I wonder if it's a rustc bug...

EPashkin:
No :(

GuillaumeGomez:
Arf... I have no idea where I could such a thing... I mean a 32 bits windows XD. I'll go around and check for old computers.

GuillaumeGomez:
Ok, I have nothing like an old 32 bits computer around me. I'll try find something further but for now, let's just merge this one...

PR #165, opened 2017-11-11
Better support for PDF, SVG and PostScript.

meh:
Don't merge yet, just need some eyes on the unsafe part.


meh:
@GuillaumeGomez this is ready as far as I'm concerned.

src/lib.rs
```diff
@@ -125,9 +122,19 @@ mod rectangle;
 mod region;
 mod surface;
 mod matrices;
+mod support;
```
EPashkin: Seems this better have `#[cfg(any(feature = "pdf", feature = "svg", feature = "ps", feature = "dox"))]`

meh: Good point, commit incoming.

EPashkin:
Second problem that all docs need moved to https://github.com/gtk-rs/lgpl-docs repo to `cairo/docs.rs`.
Current comments can be `stripped` with `rustdoc-stripper` binary from https://github.com/GuillaumeGomez/rustdoc-stripper

EPashkin:
@GuillaumeGomez there problem with regenerating stripped comments: in `ps.rs` comment moved from `impl<'a> Stream<'a>::new` to `impl<'a> Writer<'a>::new`

meh:
@EPashkin any direction to move them over without suffering too much?

EPashkin:
@meh Theoretically you just need run `rustdoc-stripper` on `cairo` repo, it produces `comments.md` that need added to `doc.rs`.
But as it not restored rightly better wait @GuillaumeGomez answer.

meh:
@EPashkin oh I see, yeah I'll wait, he's probably still traveling 🐼

GuillaumeGomez:
Ok so, if docs issue you have, regen the docs (`cargo build --features "embed-lgpl"` is my memory is correct) then add the new docs then strip everything in the comment file and send us the comment file.

If it wasn't the problem, then it just mean I need to sleep and I'll check this tomorrow. 😀

EPashkin:
@GuillaumeGomez for tomorrow check: 
I run last version of `rustdoc-stripper` for this PR,
get filled `comments.md`,
on call `rustdoc-stripper -g` it restored all comments with one exception:
in `ps.rs` comment moved from `impl<'a> Stream<'a>::new` to `impl<'a> Writer<'a>::new`

GuillaumeGomez:
That's weird. I suppose it's a `rustdoc-stripper`'s issue but I don't see what it is for the moment. I'll look into it.

EPashkin:
@meh, @GuillaumeGomez Found "solution": it restored right way if `Writer::new` also have comments

GuillaumeGomez:
That's weird...

meh:
Should I also purge the documentation from the git history?

GuillaumeGomez:
It depends if it's lgpl docs or not.

meh:
I don't know, I wrote them 🤔

GuillaumeGomez:
Then it's not lgpl. :)

meh:
I think I did something wrong while generating the docs, see pull request on docs.

cairo-sys-rs/src/enums.rs
```diff
@@ -328,6 +328,30 @@ pub enum RegionOverlap {
     Part,
 }
 
+#[cfg(any(feature = "pdf", feature = "dox"))]
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PdfVersion {
+    _1_4,
+    _1_5,
```
sdroege: See https://github.com/gtk-rs/gir/issues/480

src/ps.rs
```diff
@@ -0,0 +1,303 @@
...
+
+use ffi;
+use ffi::enums::{SurfaceType, PsLevel};
+use surface::{Surface, SurfaceExt};
+use support;
+
+#[cfg(feature = "use_glib")]
+use glib::translate::*;
+
+pub struct File {
```
sdroege: I see some potential of sharing code here between PDF and PS. It's basically the same except for the types and the functions called.

src/support.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn new(constructor: Constructor, width: f64, height: f64) -> Buffer {
+        let buffer = Box::into_raw(Box::new(Vec::new()));
+
+        Buffer {
+            buffer: buffer,
+            inner: Stream::new(constructor, width, height, move |data| {
+                unsafe {
+                    let mut out = Box::from_raw(buffer);
+                    out.extend(data);
+                    Box::into_raw(out);
```
sdroege: This seems dangerous. The buffer you store in Buffer could become invalid in theory. Box::from_raw() followed by Box::into_raw() could in theory move things to a different place on the heap.

meh: I'm pretty sure that's safe, or stuff like `owning_ref` and `rental` would explode in really funny ways.

The part I'm unsure about is the story with aliasing which might be UB there.

sdroege: The problem with from_raw/into_raw() is that they operate on owned values. They could do whatever they want with them, including moving to different memory locations. The assumption is that nothing is borrowing the data currently (and nothing is as far as the compiler is concerned, otherwise into_raw() would not compile), and with all your pointers that still point to the data you're on your own and need to ensure correctness.

I would recommend to work directly on the pointers here: `let out = &mut *buffer;`

meh: I'll look into it.

cairo-sys-rs/src/enums.rs
meh: What do I need to do exactly here? Kind of unclear from the issue.

Should I just add an `Unknown` variant?

sdroege: In the -sys crate you use an integer and constants, in the non-sys crate you have a real enum with the Unknown variant

sdroege:
What's the status/plan here?

GuillaumeGomez:
I'll see @meh this week-end. I'll try to motivate him to finish it. :)

meh:
It will happen, I believe.

cairo-sys-rs/src/enums.rs
meh: Is this still the case?

sdroege: Yes

GuillaumeGomez:
:tada: Finally! :tada:

Thanks!

yannleretaille:
:tada:

ekg:
> The Cairo library has no such concepts in its PDF, SVG, PS,... support, it just has a single object for a surface representing each format which can be constructed in various ways. Since this is a binding library, I think it's bets to stay true to the way things are organized in the original library, otherwise it gets confusing for users.

This is absolutely true. I am going to have to look for a simpler set of bindings :(

PR #173, opened 2018-01-06
deps: don't restrict c_vec version

ignatenkobrain:


GuillaumeGomez:
Why would we not want to restrict the `c_vec` version?

ignatenkobrain:
@GuillaumeGomez I think question is rather opposite, why would you? c_vec is not known (to my knowlecge) by breaking API between minor versions.

GuillaumeGomez:
Well, I have to admit that I don't care enough about it to start a debate so let's say it's ok for me. 😆 

@EPashkin: Any reason we shouldn't merge this PR? Otherwise I'm just gonna merge.

EPashkin:
IMHO restricting really unneeded, It was done for version 1.0, and just upgraded since.

GuillaumeGomez:
Then I merge.

Thanks @ignatenkobrain!

Issue #185, opened 2018-04-13
ImageSurface::create_from_png produce AV on Windows for bad input


EPashkin:
Its reason why tests `invalid_png_yields_error`, `io_error_yields_cairo_read_error` causes failure in appveyor

EPashkin:
Seems something in libcairo was updated in bad way.
Tests added in https://github.com/gtk-rs/cairo/pull/146 by @federicomenaquintero and currently it fails.
But on https://ci.appveyor.com/project/GuillaumeGomez/cairo-p0df1/build/1.0.329 it was passed

federicomenaquintero:
If you narrow this down to a bug in Cairo upstream, could you please tell me about it? (Or, @GuillaumeGomez we can talk about it this week at the hackfest?)  I'm trying to fix things in Cairo upstream, so maybe I can handle this if I know what it's about :)

EPashkin:
New cairo 1.15.12 don't panics, seems it was attempt optimizing usage `setjmp` in 1.15.10, that was reverted.
I tried restart appveyor for https://github.com/gtk-rs/cairo/pull/184 as final test, but it can't download rustup now.

EPashkin:
Appveyor still used cairo-1.15.10-1

EPashkin:
Added https://github.com/gtk-rs/cairo/pull/186 that use update pacman before installing gtk, it solved issue

PR #187, opened 2018-04-22
Add more checks

GuillaumeGomez:


GuillaumeGomez:
I found the issue here on osx: the method `ImageSurface::create_from_png`. It fails when leaving the `read_func`. And that's pretty much all we can get at this point... I suppose there is a bug in the osx implementation. We can still make a workaround for osx.

EPashkin:
It not related to #185?

GuillaumeGomez:
Ah maybe?

GuillaumeGomez:
It seems to be working now so I merge (but the issue on OSX and potentially windows remains).

PR #188, opened 2018-04-26
Add to_glib_none_mut for context

etrombly:
Just modified the implementation from Gdk Rectangle, so not sure if it's safe for a context. It's working so far in my application without issues.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -54,6 +54,16 @@ impl<'a> ToGlibPtr<'a, *mut ffi::cairo_t> for &'a Context {
 }
 
 #[cfg(feature = "use_glib")]
+impl<'a> ToGlibPtrMut<'a, *mut ffi::cairo_t> for Context {
+    type Storage = &'a mut Context;
+
+    #[inline]
+    fn to_glib_none_mut(&'a mut self) -> StashMut<'a, *mut ffi::cairo_t, Self> {
+        StashMut(self.0 as *mut ffi::cairo_t, self)
+    }
+}
+
+#[cfg(feature = "use_glib")]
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Seems good to me.

src/context.rs
```diff
@@ -53,6 +53,16 @@ impl<'a> ToGlibPtr<'a, *mut ffi::cairo_t> for &'a Context {
     }
 }
 
+#[cfg(feature = "use_glib")]
+impl<'a> ToGlibPtrMut<'a, *mut ffi::cairo_t> for Context {
+    type Storage = &'a mut Context;
+
+    #[inline]
+    fn to_glib_none_mut(&'a mut self) -> StashMut<'a, *mut ffi::cairo_t, Self> {
```
EPashkin: If this needed for manual code, better use http://gtk-rs.org/docs/glib/translate/fn.mut_override.html

etrombly: It's for bindings generated by gir. https://github.com/etrombly/osmgpsmap-rs

etrombly: Thanks, that works well. Also since you know this well, map_point.rs line 46, is generating a f32 for an argument and it expects a *mut f32. How do I fix it with gir?

You can close this pull request if to_glib_none_mut shouldn't be used.

EPashkin: It must be looks like `get_degrees`.
Seems `OsmGpsMap-1.0.gir` has error and `osm_gps_map_point_get_radians` need fix too looks like `osm_gps_map_point_get_degrees`.
Or this function can be implemented in manual way.


And please close issue yourself when no problem remains.

EPashkin: Also `&mut self` for all getters better fixed with `const=true` too

PR #191, opened 2018-06-10
cairo-sys-rs: Don't leak system library path information

philn:
Otherwise the system library path takes precedence over paths that might be
specified by other dependencies of a crate depending on cairo and thus introduce
unrelated link errors.


philn:
The CI failure seems unrelated (futures).

GuillaumeGomez:
Thanks!

PR #197, opened 2018-07-11
Add svg surface

jsparber:

PR #143, opened 2017-07-15
Added PDF as a target Surface

savage13:
This was modeled after ImageSurface.  It is missing the rest of the PDF Surface functions at https://www.cairographics.org/manual/cairo-PDF-Surfaces.html

I am unsure if the conversion to ToGlibPtr functions work properly.  

I would image Postscript would be similar.

PR #198, opened 2018-07-11
add SVGSurface fix #196

### LINKED ISSUE #196: Add SVGSurface

jsparber:


EPashkin:
Please "Fix #197" in first comment here too

EPashkin:
* not in git comment,
in github comment that says `No description provided.` now

EPashkin:
Hm, this is enought too. Thanks
👍 After CI pass

EPashkin:
cc @GuillaumeGomez. @sdroege 
CI partially pass, but not completed now

src/svg_surface.rs
```diff
@@ -0,0 +1,159 @@
+// Copyright 2015-2016, The Gtk-rs Project Developers.
```
sdroege: We're in 2018 nowadays :)

src/svg_surface.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use surface::{Surface, SurfaceExt};
+
+#[cfg(feature = "use_glib")]
+use glib::translate::*;
+
+pub struct SVGSurface(Surface);
+
+extern "C" {
+    pub fn cairo_svg_surface_create (filename: *const c_char,
```
sdroege: Shouldn't this be in the sys crate?

src/svg_surface.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub unsafe fn from_raw_full(ptr: *mut ffi::cairo_surface_t) -> SVGSurface {
+        Self::from(Surface::from_raw_full(ptr)).unwrap()
+    }
+
+    pub fn create<T: AsRef<Path>>(filename: T, width: f64, height: f64) -> SVGSurface
+    {
+        // Convert: AsRef<Path> -> Cow<str> -> str
+        let s = filename.as_ref().to_string_lossy().into_owned();
+        let file = CString::new(s).unwrap();
+        unsafe { Self::from_raw_full(cairo_svg_surface_create(file.as_ptr(), width, height)) }
```
sdroege: This can probably fail, e.g. if the filename is unusable?

src/svg_surface.rs
jsparber: I used the same code as PDFSurface

src/svg_surface.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    #[doc(hidden)]
+    pub unsafe fn from_raw_full(ptr: *mut ffi::cairo_surface_t) -> SVGSurface {
+        Self::from(Surface::from_raw_full(ptr)).unwrap()
+    }
+
+    pub fn create<T: AsRef<Path>>(filename: T, width: f64, height: f64) -> SVGSurface
+    {
+        // Convert: AsRef<Path> -> Cow<str> -> str
+        let s = filename.as_ref().to_string_lossy().into_owned();
```
sdroege: Lossy does not seem ideal. Also does Cairo always expect UTF8 on all platforms? If not this should go through `OsStr`. `Path` can be converted to `OsStr`.

src/svg_surface.rs
```diff
@@ -0,0 +1,159 @@
...
+
+pub struct SVGSurface(Surface);
+
+extern "C" {
+    pub fn cairo_svg_surface_create (filename: *const c_char,
+                                     width_in_points: c_double,
+                                     height_in_points: c_double) -> *mut ffi::cairo_surface_t;
+}
+
+impl SVGSurface {
```
sdroege: A binding for `cairo_svg_surface_restrict_to_version()` might be useful

src/svg_surface.rs
sdroege: Seems equally weird :) I meant in here btw https://github.com/gtk-rs/cairo/tree/master/cairo-sys-rs

EPashkin: Yes, but we already allowed this in #143,
so IMHO we can ignore this, and fix with PDF yourself

jsparber: would it make any problems to move this code to the sys?

sdroege: I don't think so, why would it?

jsparber: No idea

src/svg_surface.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    #[doc(hidden)]
+    pub unsafe fn from_raw_full(ptr: *mut ffi::cairo_surface_t) -> SVGSurface {
+        Self::from(Surface::from_raw_full(ptr)).unwrap()
+    }
+
+    pub fn create<T: AsRef<Path>>(filename: T, width: f64, height: f64) -> SVGSurface
+    {
+        unsafe { Self::from_raw_full(cairo_svg_surface_create(filename.as_ref().to_glib_none().0, width, height)) }
```
sdroege: This would require that the Cairo filename convention is the same as GLib's. Are they? That is, arbitrary bytes on UNIX, UTF8 on Windows?

And unfortunately `to_glib_none()` is only available if the cairo bindings are compiled with GLib support

sdroege: The docs say
```
This function always returns a valid pointer, but it will return a pointer to a "nil" surface if an error such as out of memory occurs. You can use cairo_surface_status() to check for this.
```

Should we check for this here and elsewhere and directly return an `Err`?

src/svg_surface.rs
jsparber: Ok tired to move them with out any problem, I will update it after lunch

jsparber: i updated it.

src/svg_surface.rs
jsparber: i reverted the changes I made before, i'm not really sure how to address this

src/svg_surface.rs
jsparber: i'm not really sure how to fix this

src/svg_surface.rs
sdroege: You'll have to check the documentation or the code. It's kind of important to convert correctly or otherwise the function will fail in certain circumstances.

jsparber:
I don't have a lot of time right now, therefore can we merge this and then fix the error handling for PDFSurface and SVGSurface?

sdroege:
It's not just error handling but also the path conversions. Please file issues about that then and we should block the release on the resolution of the path handling it least.

src/svg_surface.rs
```diff
@@ -0,0 +1,167 @@
...
+        { // String
+            let filename = output.join("test3.svg");
+            let s = match filename.to_str() {
+                None => panic!("Error converting Path to String"),
+                Some(x) => String::from(x),
+            };
+            let surface = SVGSurface::create(s, 100., 100.);
+            let cr = Context::new(&surface);
+            draw_x(&cr);
+        }
```
EPashkin: @sdroege You don't think that many cases in this test actually try to test
rustc ability to generate call code? IMHO we need only one test without units and one with.

sdroege: I agree. But what could make sense is to add tests that actually check that the path conversion works correctly and the resulting files actually exist with their correct names.

Create some files with non-ASCII names and check with the API from `std` if those files actually exist with those names in the end. That would also solve my concern about the conversion as we run tests on Linux and Windows, and those are the platforms with the main differences.

sdroege:
@GuillaumeGomez @EPashkin How should we go ahead with this? Does one of you want to update this PR so it can be merged?

GuillaumeGomez:
If @EPashkin doesn't want to, I'll I suppose.

jsparber:
I guess we can close this because of https://github.com/gtk-rs/cairo/commit/d39b674ef9e7a73cafebf7fa32e4eb7ae82f6900

GuillaumeGomez:
Ok, thanks anyway!

PR #200, opened 2018-07-14
remove versions from git dependencies

vhdirk:
Again: same as gtk-rs/pangocairo#20
I chose to do it a little different in this one because of the optionality of glib, and to keep it similarly looking.


GuillaumeGomez:
Similar looking is good too.

PR #204, opened 2018-08-15
Allow any AsMut<[u8]> + Send + 'static to be used for creating an Ima…

sdroege:
…geSurface


sdroege:
Ok, I'll do that once this here is merged. Just a matter of removing that parameter

sdroege:
Ugh actually not that simple, the code assumes this weird free-function. I refer to https://github.com/gtk-rs/examples/issues/181, that code needs some serious refactoring.

I'll fix it up

PR #217, opened 2018-11-20
Enum rework

GuillaumeGomez:
cc @sdroege @EPashkin


EPashkin:
Looks good, but `cairo_content_t` seems seems still needed (or `CairoContent` renamed back)
In this way cairo-sys don't need glib dependency in Cargo.toml?

GuillaumeGomez:
I've put back all `*_t` aliases.

EPashkin:
Maybe leave enums in sys with values otherwise sys crate will be less usable?

GuillaumeGomez:
I prefer to move enums outside of it and just keep aliases. It's more consistent with other gtk-rs crates this way.

GuillaumeGomez:
(And for failing crates, I'll have to update the corresponding parts by hand.)

EPashkin:
Not all other gtk-rs crates have enums both in sys and normal.

GuillaumeGomez:
True but the missing `__Unknown` field really bothered me. What do you think of it @sdroege?

EPashkin:
Usually we have `#repr[C] enum` in sys and rust enum with `__Unknown` in main crate.

EPashkin:
.. as you already created rust enums to main you can just restore enums in sys and IMHO it will be good.

GuillaumeGomez:
Which crate has enums in sys? Didn't find one yet (looked into gtk and gdk only for the moment...)

EPashkin:
@GuillaumeGomez Yes, I was wrong.
We use `c_int` type with constants instead enum in sys.

EPashkin:
.. but do it in cairo-rs-sys will be IMHO too bothersome.

EPashkin:
@GuillaumeGomez Thanks.
IMHO only last is remove `[dependencies.glib]` and `[dependencies.gobject-sys]` in sys.

GuillaumeGomez:
I'll check if nothing needs them in `cairo-sys` and if not, I'll remove them.

src/context.rs
```diff
@@ -10,12 +10,17 @@ use std::ops;
 use ::paths::Path;
 use ::font::{TextExtents, TextCluster, FontExtents, ScaledFont, FontOptions, FontFace, Glyph};
 use ::matrices::{Matrix, MatrixTrait};
-use ffi::enums::{
+use ::enums::{
```
sdroege: Or directly use the re-exported enums from the crate root?

src/lib.rs
```diff
@@ -88,6 +87,8 @@ pub use enums::{
     RegionOverlap,
     SurfaceType,
 };
+#[cfg(any(feature = "v1_12", feature = "dox"))]
+pub use enums::MeshCorner;
```
sdroege: Or just `pub use enums::*;`?

sdroege:
>  IMHO only last is remove `[dependencies.glib]` and `[dependencies.gobject-sys]` in sys.

`glib-sys` is still needed for all the `get_type()` functions in the -sys crate at least. `gobject-sys` might be unneeded

PR #219, opened 2018-11-23
Add cairo_font_type_t

EPashkin:
Part of https://github.com/gtk-rs/sys/issues/116

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -34,46 +34,37 @@ use libc::{c_void, c_int, c_uint, c_char, c_uchar, c_double, c_ulong};
 #[cfg(any(feature = "xlib", feature = "dox"))]
 use x11::xlib;
 
-pub type CairoAntialias = c_int;
-pub type CairoContent = c_int;
-pub type CairoExtend = c_int;
-pub type CairoFillRule = c_int;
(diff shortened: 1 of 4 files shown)


EPashkin:
Maybe all enums need be exported under cairo names?

cairo-sys-rs/src/lib.rs
```diff
@@ -58,6 +58,7 @@ pub type CairoSvgUnit = c_int;
 pub type CairoTextClusterFlags = c_int;
 
 pub type cairo_antialias_t = CairoAntialias;
+pub type cairo_font_type_t = CairoFontType;
```
sdroege: Maybe in -sys we should only have the names as in C, and the CamelCase enum types only in the non-sys crate? That's more consistent with everything else.

EPashkin: Agree.
@GuillaumeGomez What you think?

GuillaumeGomez: I agree as well.

EPashkin:
Ok then I do changes here

EPashkin:
Updated
Also found that we have two cairo_fill_rule_t :wink:

GuillaumeGomez:
Thanks a lot!

PR #220, opened 2018-11-24
Add get_mime_data(), set_mime_data() and supports_mime_type()  to cario surface

jsparber:


src/surface.rs
```diff
@@ -47,6 +47,61 @@ impl Surface {
                                                                   height))
         }
     }
+
+    pub fn get_mime_data(&self, mime_type: &str) -> Option<&[u8]> {
```
sdroege: This is unsafe: nothing is preventing the caller to set a different data for the mimetype after calling this function, and then using the (now freed) byte array of the previously set data

src/surface.rs
```diff
@@ -47,6 +47,61 @@ impl Surface {
...
+
+        let status = unsafe {
+            ffi::cairo_surface_set_mime_data(
+                self.to_raw_none(),
+                mime_type.to_glib_none().0,
+                data,
+                size as u64,
+                Some(unbox::<T>),
+                user_data as *mut _,
+                )
```
EPashkin: Please make indent smaller

src/surface.rs
jsparber: Should i make a copy of the data and return a vector?

sdroege: That would be one option, yes. Alternative would be something like the data getter from the image surface but that's very limiting (requires refcount of 1).

src/surface.rs
jsparber: How wide should the indent be? Isn't it 4 spaces?

src/surface.rs
jsparber: What would be the advantage of using a getter similar to the that one in image surface?

sdroege: No copying of the data into a `Vec` :)

jsparber: Oh I thought something more than that. I used to_vec()

sdroege: Great, thanks :)

src/surface.rs
sdroege: 4 spaces, yes

src/surface.rs
```diff
@@ -47,6 +47,60 @@ impl Surface {
...
+                ).to_vec())
+            } else {
+                None
+            }
+        }
+    }
+
+    pub fn set_mime_data<T: AsRef<[u8]> + Send + 'static>(&self,
+                                                          mime_type: &str,
+                                                          slice: T) -> Result<(), Status> {
```
sdroege: This one should probably be indented a bit differntly:

```rust
    pub fn set_mime_data<T: AsRef<[u8]> + Send + 'static>(
        &self,
        mime_type: &str,
        slice: T) -> Result<(), Status>
```

Or something like that. Indent logically, not visually :)

src/surface.rs
```diff
@@ -47,6 +47,60 @@ impl Surface {
...
+        let user_data = Box::into_raw(b);
+
+        let status = unsafe {
+            ffi::cairo_surface_set_mime_data(
+                self.to_raw_none(),
+                mime_type.to_glib_none().0,
+                data,
+                size as u64,
+                Some(unbox::<T>),
+                user_data as *mut _)
```
sdroege: The closing parenthesis should be on the next line at the same level as `ffi::`

src/surface.rs
sdroege: That's why we really need rustfmt ;)

jsparber: i really want rustfmt.

jsparber:
what should i use instead of to_glib_none() and ptr::null_mut(), since the CI fails because of them?

jsparber:
I think i fixed all indentation issues, but we really need rustfmt

sdroege:
> what should i use instead of to_glib_none() and ptr::null_mut(), since the CI fails because of them?

The former is only available if cairo is built with glib support. You should use `CStr` / `CString` for strings, or where are the problems?

For `ptr`, that should work. How does it fail?

jsparber:
I didn't include ptr in surface, I guess cagro just got it from some where else. Anyhow it is fixed now. Also the CString thing.

jsparber:
@sdroege std::ptr is included already inside the gvalue_impl macro, but this is used only when glib is enabled. How should I import in surface.rs?

sdroege:
> @sdroege std::ptr is included already inside the gvalue_impl macro, but this is used only when glib is enabled therefore the build fails if glib is disabled. How should I import in surface.rs?

Just add a use statement for it? :)

jsparber:
> > @sdroege std::ptr is included already inside the gvalue_impl macro, but this is used only when glib is enabled therefore the build fails if glib is disabled. How should I import in surface.rs?
> 
> Just add a use statement for it? :)

Rust doesn't like to include it twice. I now added an annotation to the use statement to include std::ptr only when glib is disabled.

sdroege:
> Rust doesn't like to include it twice. I now added an annotation to the use statement to include std::ptr only when glib is disabled.

You could remove the use statement from the macro (and use it as `::std::ptr::...`).

src/surface.rs
EPashkin: First comment was about closing bracket has same indent as parameters.

jsparber:
> > Rust doesn't like to include it twice. I now added an annotation to the use statement to include std::ptr only when glib is disabled.
> 
> You could remove the use statement from the macro (and use it as `::std::ptr::...`).

i tried std::ptr before (which was wrong). let's hope now it works

src/surface.rs
jsparber: I hope you are happy with the last change @EPashkin ;)

EPashkin: Yes, thanks 😉

src/lib.rs
```diff
@@ -141,6 +140,7 @@ pub use xcb::{
 };
 
 pub mod prelude;
+pub mod constants;
```
sdroege: Maybe
```rust
mod constants;
pub use constants::*;
```
instead?

sdroege:
Looks good to me, but you might want to add an unsafe getter for the data too that does not copy. In librsvg you can ensure that the data stays valid all the time, and could prevent expensive copies that way.

jsparber:
just a unsave function which returns option<&[u8]>? Any ideas how i should name it? I thought about get_mime_data_raw()

src/lib.rs
jsparber: Out of curiosity whats the difference between those two? Yours looks cleaner ;)

sdroege: Yours requires users to do `cairo::constants::BLABLA`, mine would be `cairo::BLABLA`

sdroege:
> just a unsave function which returns option<&[u8]>? Any ideas how i should name it? I thought about get_mime_data_raw()

Sounds good to me

sdroege:
Seems good to me now

GuillaumeGomez:
Thanks!

PR #233, opened 2019-01-23
Make Surface methods &self from &mut self, remove Send impls

### LINKED ISSUE #232: Shouldn't ImageSurface has an AsMut<Surface> impl?
Currently I can get a `&Surface` from a `&ImageSurface`, but I can't get a `&mut Surface` from `&mut ImageSurface` because there isn't an impl `AsMut<Surface> for ImageSurface`. This makes some operations impossible, e.g. I can't call `set_device_offset()` (added with #230) on an `ImageSurface`.

osa1:
Fixes #232

### DIFF
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -45,9 +45,9 @@ impl ImageSurface {
         }
     }
 
-    pub fn create_for_data<D: AsMut<[u8]> + Send + 'static>(data: D, format: Format, width: i32, height: i32,
+    pub fn create_for_data<D: AsMut<[u8]> + 'static>(data: D, format: Format, width: i32, height: i32,
                               stride: i32) -> Result<ImageSurface, Status> {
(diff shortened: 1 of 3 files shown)


sdroege:
:+1:

GuillaumeGomez:
:+1:

sdroege:
Still looks good to me :)

sdroege:
@GuillaumeGomez This of course breaks the example that makes use of the `Send` impl on the image surface but that example should be rewritten IMHO anyway, it's quite ugly :) I can make it work again without rewriting it all though, once this is merged here.

GuillaumeGomez:
Ok, then I'll just merge and wait for your example rewrite. :)

Thanks @osa1 !

sdroege:
> Ok, then I'll just merge and wait for your example rewrite. :)

Can you create an issue for that and CC me there so I don't forget?

GuillaumeGomez:
Sure.

PR #234, opened 2019-01-23
Refactor interface in PDF

vojtechkral:
Hi,
  this is a follow-up to the discussion in d39b674ef9e7a73cafebf7fa32e4eb7ae82f6900
This PR is not meant to be merged as-is, but as a place to discuss the interface of the additional surfaces (PDF, PS, SVG, ...).

I did the refactoring on PDF so far as an example (the others are analogous I believe).

I tried my best to reconcile the requirements mentioned in the talk under the commit, let me know if I forgot something.
Hopefully none of the changes are too offensive.

Basically there's the basic `File` type, I think that one is unchaged as far as public interface is concerned. Then there's `Writer` and `RefWriter`, the naming is inspired by `Cell` and `RefCell` but is open for bikeshedding - let me know if you prefer different naming. These two additional surface types `Deref` to the `File`. `RefWriter` is basically the same as the former `Writer` and `Writer` is the same thing except taking the ownership of the `io::Write` thing passed in.

In the end I decided to leave the `W: io::Write` type as part of the `Writer` and `RefWriter` types, this allows for single boxing and no boxing respectively and also allows to get the `W` back from the `Writer` using `finish()`.

Please see the tests to get a feeling of how the interface is used.

cc @meh @sdroege

PR #236, opened 2019-02-06
Finalize PDF, SVG, PS work from #234

vojtechkral:
Finalized the refactor and added remaining bindings for those three surface types.

Two things I'm not sure about:
- Noticed copyrights were off, fixed it, hopefully correctly.
- `add_outline()` in PDF uses bit flags. I just added integer constants that can be or'd. I thought bringing in the bitflags crate might not be appropriate.

Sorry for such a large diff.

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -58,6 +58,10 @@ pub type cairo_svg_unit_t = c_int;
 pub type cairo_text_cluster_flags_t = c_int;
 
 #[cfg(any(feature = "pdf", feature = "dox"))]
+pub type cairo_pdf_outline_flags_t = c_int;
+#[cfg(any(feature = "pdf", feature = "dox"))]
+pub type cairo_pdf_metadata_t = c_int;
+#[cfg(any(feature = "pdf", feature = "dox"))]
 pub type cairo_pdf_version_t = c_int;
 #[cfg(any(feature = "svg", feature = "dox"))]
 pub type cairo_svg_version_t = c_int;
@@ -258,6 +262,12 @@ pub struct cairo_bool_t{
 impl cairo_bool_t {
     pub fn as_bool(self) -> bool {
         self.value != 0
+    }
+}
+
+impl From<bool> for cairo_bool_t {
+    fn from(b: bool) -> cairo_bool_t {
+        cairo_bool_t { value: b as _ }
     }
 }
 
@@ -593,9 +603,31 @@ extern "C" {
     #[cfg(any(feature = "pdf", feature = "dox"))]
     pub fn cairo_pdf_surface_restrict_to_version (surface: *mut cairo_surface_t, version: cairo_pdf_version_t);
     #[cfg(any(feature = "pdf", feature = "dox"))]
+    pub fn cairo_pdf_get_versions (versions: *mut *mut cairo_pdf_version_t,
+                                   num_versions: *mut c_int);
+    #[cfg(any(feature = "pdf", feature = "dox"))]
+    pub fn cairo_pdf_version_to_string (version: cairo_pdf_version_t) -> *const c_char;
+    #[cfg(any(feature = "pdf", feature = "dox"))]
     pub fn cairo_pdf_surface_set_size (surface: *mut cairo_surface_t,
                                       width_in_points: f64,
                                       height_in_points: f64);
+    #[cfg(any(feature = "pdf", feature = "dox"))]
+    pub fn cairo_pdf_surface_add_outline (surface: *mut cairo_surface_t,
+                                          parent_id: c_int,
+                                          utf8: *const c_char,
+                                          link_attribs: *const c_char,
+                                          flags: cairo_pdf_outline_flags_t) -> c_int;
+    #[cfg(any(feature = "pdf", feature = "dox"))]
+    pub fn cairo_pdf_surface_set_metadata (surface: *mut cairo_surface_t,
+                                           metadata: cairo_pdf_metadata_t,
+                                           utf8: *const c_char);
+    #[cfg(any(feature = "pdf", feature = "dox"))]
+    pub fn cairo_pdf_surface_set_page_label (surface: *mut cairo_surface_t,
+                                             utf8: *const c_char);
+    #[cfg(any(feature = "pdf", feature = "dox"))]
+    pub fn cairo_pdf_surface_set_thumbnail_size (surface: *mut cairo_surface_t,
+                                                 width: c_int,
+                                                 height: c_int);
 
     // CAIRO SVG
     #[cfg(any(feature = "svg", feature = "dox"))]
@@ -613,6 +645,12 @@ extern "C" {
     pub fn cairo_svg_surface_get_document_unit(surface: *const cairo_surface_t) -> cairo_svg_unit_t;
     #[cfg(any(feature = "svg", feature = "dox"))]
     pub fn cairo_svg_surface_set_document_unit(surface: *mut cairo_surface_t, unit: cairo_svg_unit_t);
+    #[cfg(any(feature = "svg", feature = "dox"))]
+    pub fn cairo_svg_get_versions (versions: *mut *mut cairo_svg_version_t,
+                                   num_versions: *mut c_int);
+    #[cfg(any(feature = "svg", feature = "dox"))]
+    pub fn cairo_svg_version_to_string (version: cairo_svg_version_t) -> *const c_char;
+
     // CAIRO PS
     #[cfg(any(feature = "ps", feature = "dox"))]
     pub fn cairo_ps_surface_create (filename: *const c_char,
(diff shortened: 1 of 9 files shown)


sdroege:
@meh please review if this breaks your things :)

cairo-sys-rs/src/lib.rs
```diff
@@ -261,6 +265,12 @@ impl cairo_bool_t {
     }
 }
 
+impl From<bool> for cairo_bool_t {
+    fn from(b: bool) -> cairo_bool_t {
+        cairo_bool_t { value: b as _ }
```
sdroege: A `match` might be better, I don't think it's guaranteed that they're `true = 1`, `false = 0`.

src/constants.rs
```diff
@@ -14,3 +17,8 @@ pub const MIME_TYPE_CCITT_FAX: &str = "image/g3fax";
 pub const MIME_TYPE_CCITT_FAX_PARAMS: &str = "application/x-cairo.ccitt.params";
 pub const MIME_TYPE_EPS: &str = "application/postscript";
 pub const MIME_TYPE_EPS_PARAMS: &str = "application/x-cairo.eps.params";
+
+pub const PDF_OUTLINE_ROOT: i32 = 0;
+pub const PDF_OUTLINE_OPEN: i32 = ffi::PDF_OUTLINE_FLAG_OPEN;
+pub const PDF_OUTLINE_BOLD: i32 = ffi::PDF_OUTLINE_FLAG_BOLD;
+pub const PDF_OUTLINE_ITALIC: i32 = ffi::PDF_OUTLINE_FLAG_ITALIC;
```
sdroege: Better to use `bitflags` for this

src/pdf.rs
```diff
@@ -1,21 +1,42 @@
...
 
 #[cfg(feature = "use_glib")]
 use glib::translate::*;
 
 
+pub fn get_versions() -> Vec<PdfVersion> {
+    let vers_slice = unsafe {
+        let mut vers_ptr: *mut ffi::cairo_pdf_version_t = mem::uninitialized();
+        let mut num_vers = 0;
+        ffi::cairo_pdf_get_versions(&mut vers_ptr as _, &mut num_vers as _);
```
sdroege: Might this return `0` or `NULL`?

src/pdf.rs
```diff
@@ -48,6 +69,36 @@ impl File {
             ffi::cairo_pdf_surface_set_size(self.inner.to_raw_none(), width, height);
         }
     }
+
+    pub fn set_metadata(&self, metadata: PdfMetadata, value: &str) {
+        unsafe {
+            ffi::cairo_pdf_surface_set_metadata(self.inner.to_raw_none(), metadata.into(), value.as_ptr() as _);
```
sdroege: This does not work (in general). `&str` is not `\0`-terminated so you can't just use `as_ptr()` on it like this.

There's `CStr` / `CString` for working with this correctly.

src/ps.rs
```diff
@@ -1,24 +1,52 @@
...
 
 #[cfg(feature = "use_glib")]
 use glib::translate::*;
 
+
+pub fn get_levels() -> Vec<PsLevel> {
+    let lvls_slice = unsafe {
+        let mut vers_ptr: *mut ffi::cairo_ps_level_t = mem::uninitialized();
+        let mut num_vers = 0;
+        ffi::cairo_ps_get_levels(&mut vers_ptr as _, &mut num_vers as _);
```
sdroege: Can this return `0` / `NULL`?

src/svg.rs
```diff
@@ -1,69 +1,79 @@
...
-            unsafe {
-                SvgUnit::from(ffi::cairo_svg_surface_get_document_unit(self.inner.to_raw_none()))
-            }
-        }
+
+pub fn get_versions() -> Vec<SvgVersion> {
+    let vers_slice = unsafe {
+        let mut vers_ptr: *mut ffi::cairo_svg_version_t = mem::uninitialized();
+        let mut num_vers = 0;
+        ffi::cairo_svg_get_versions(&mut vers_ptr as _, &mut num_vers as _);
```
sdroege: And how about this? :)

src/pdf.rs
sdroege: Please check the whole cairo codebase for this while you're at it, and also especially the code of your previews PR :)

cairo-sys-rs/src/lib.rs
GuillaumeGomez: Or `b.to_glib()` maybe? :)

sdroege: No, glib support in cairo is optional.

GuillaumeGomez: Ah crap, forgot that...

vojtechkral: Ok... I'd use an `if` here if that's ok with you...

src/constants.rs
vojtechkral: Allright, will do...

src/pdf.rs
vojtechkral: I don't know, the documentation says nothing on that and it doesn't say anything about the returned pointer memory management either. The implementation simply returns a pointer to an internal static array (see [here](https://github.molgen.mpg.de/git-mirror/cairo/blob/master/src/cairo-pdf-surface.c#L165)). To be in line with that, it would probably be better to return a `&'static str` instead of `String` here. (The same applies to the others.)

src/pdf.rs
vojtechkral: Ok

cairo-sys-rs/src/lib.rs
sdroege: Sure, either way is fine :)

src/pdf.rs
sdroege: Ok so can't be `NULL` then. Thanks for checking!

For making it a `&str` you need to count until the `\0`-terminator for each string and then use the corresponding functions, but that would indeed be nicer.

vojtechkral:
Updated

meh:
I have some slight concerns still with the removal of the `Buffer` crap, but I think it's doable with `Writer::finish` anyway, so whatever.

sdroege:
> I have some slight concerns still with the removal of the `Buffer` crap, but I think it's doable with `Writer::finish` anyway, so whatever.

So does it work for your use-case or is more work needed? :)

src/pdf.rs
```diff
@@ -29,10 +29,10 @@ pub fn get_versions() -> Vec<PdfVersion> {
     vers_slice.iter().map(|v| PdfVersion::from(*v)).collect()
 }
 
-pub fn version_to_string(version: PdfVersion) -> Option<String> {
+pub fn version_to_string(version: PdfVersion) -> Option<&'static str> {
     unsafe {
         let res = ffi::cairo_pdf_version_to_string(version.into());
-        res.as_ref().and_then(|cstr| CStr::from_ptr(cstr as _).to_str().ok()).map(String::from)
+        res.as_ref().and_then(|cstr| CStr::from_ptr(cstr as _).to_str().ok())
```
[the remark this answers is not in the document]
vojtechkral: It [does](https://github.molgen.mpg.de/git-mirror/cairo/blob/master/src/cairo-pdf-surface.c#L666) return `NULL` if the `version` argument is not one of the defined values, which the bindings in `enum.rs` generally don't protect against (ie. there's the `__Unknown(i32)` variant). I suppose the enum could be made exhaustive and there could be an assert in it's from From<ffi> implementation. Then this could simplified. I don't have a strong feeling one way or the other, feel free to pick whichever you think is right...

vojtechkral:
@meh 

> I have some slight concerns still with the removal of the `Buffer` crap, but I think it's doable with `Writer::finish` anyway, so whatever.

I'd like to improve the ergonomics if I can. Feel free to post usage example of `Buffer` vs `Writer` or something. Would an ability to reference the `io::Write` object while the `Writer` is still in scope help you? I had been thinking about adding that one anyway...

src/pdf.rs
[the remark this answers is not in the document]
sdroege: It even mentions in the docs of that function that it can return `NULL`, so this is fine then.

meh:
> > I have some slight concerns still with the removal of the `Buffer` crap, but I think it's doable with `Writer::finish` anyway, so whatever.
> 
> So does it work for your use-case or is more work needed? :)

It works for the current use case, my only concern is when we need to start sending PDFs through HTTP responses and such.

I honestly don't remember why I wanted to be able to access a `pdf::Buffer` through `AsRef<[u8]>`, so using `finish()` and sending the buffer through might be enough.

Only time will tell.

sdroege:
@meh what you're missing is a way to incrementally write things instead of having the whole thing in memory at once?

vojtechkral:
I added the `AsRef<[u8]>` and `writer()` + `writer_mut()` for convenience.

sdroege:
@GuillaumeGomez Let get this in then?

GuillaumeGomez:
Thanks @vojtechkral !

vojtechkral:
Thank you all for bearing with me through this :)

PR #241, opened 2019-02-13
Add missing 1.12 items

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -102,12 +102,6 @@ pub struct cairo_pattern_t(u8);
 
 #[cfg(any(feature = "xcb", feature = "dox"))]
 #[repr(C)]
-pub struct cairo_device_t(c_void);
-#[cfg(any(feature = "xcb", feature = "dox"))]
-debug_impl!(cairo_device_t);
-
-#[cfg(any(feature = "xcb", feature = "dox"))]
-#[repr(C)]
 pub struct xcb_connection_t(c_void);
 #[cfg(any(feature = "xcb", feature = "dox"))]
 debug_impl!(xcb_connection_t);
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -592,6 +592,8 @@ extern "C" {
     pub fn cairo_surface_get_fallback_resolution(surface: *mut cairo_surface_t, x_pixels_per_inch: *mut c_double, y_pixels_per_inch: *mut c_double);
     pub fn cairo_surface_set_fallback_resolution(surface: *mut cairo_surface_t, x_pixels_per_inch: c_double, x_pixels_per_inch: c_double);
     pub fn cairo_recording_surface_get_extents(surface: *mut cairo_surface_t, extents: *mut cairo_rectangle_t) -> cairo_bool_t;
+    pub fn cairo_recording_surface_create(content: cairo_content_t, extents: *const cairo_rectangle_t) -> *mut cairo_surface_t;
+    pub fn cairo_recording_surface_ink_extents(surface: *mut cairo_surface_t, x0: *mut c_double, y0: *mut c_double, width: *mut c_double, height: *mut c_double);
     pub fn cairo_surface_create_similar_image(other: *mut cairo_surface_t, format: cairo_format_t, width: c_int, height: c_int) -> *mut cairo_surface_t;
     pub fn cairo_surface_map_to_image(surface: *mut cairo_surface_t, extents: *const cairo_rectangle_int_t) -> *mut cairo_surface_t;
     pub fn cairo_surface_unmap_image(surface: *mut cairo_surface_t, image: *mut cairo_surface_t);
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -97,8 +97,7 @@ pub struct cairo_device_t(c_void);
 debug_impl!(cairo_device_t);
 
 #[repr(C)]
-#[derive(Copy,Clone,Debug)]
-pub struct cairo_pattern_t(u8);
+pub struct cairo_pattern_t(c_void);
 
 #[cfg(any(feature = "xcb", feature = "dox"))]
 #[repr(C)]
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -36,6 +36,7 @@ use x11::xlib;
 
 pub type cairo_antialias_t = c_int;
 pub type cairo_content_t = c_int;
+pub type cairo_device_type_t = c_int;
 pub type cairo_extend_t = c_int;
 pub type cairo_fill_rule_t = c_int;
 pub type cairo_filter_t = c_int;
@@ -43,6 +44,7 @@ pub type cairo_font_slant_t = c_int;
 pub type cairo_font_type_t = c_int;
 pub type cairo_font_weight_t = c_int;
 pub type cairo_format_t = c_int;
+pub type cairo_ft_synthesize_t = c_uint;
 pub type cairo_hint_metrics_t = c_int;
 pub type cairo_hint_style_t = c_int;
 pub type cairo_line_cap_t = c_int;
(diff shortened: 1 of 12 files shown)


sdroege:
You mean 1.12?

GuillaumeGomez:
... Indeed.

src/device.rs
```diff
@@ -0,0 +1,265 @@
...
+                None
+            } else {
+                Some(Self::from_raw_full(p))
+            }
+        }
+    }
+
+    pub fn from_recording_surface(&self, surface: &Surface) -> Status {
+        unsafe {
+            Status::from(ffi::cairo_script_from_recording_surface(self.to_raw_none(),
```
sdroege: You need to ensure that this is actually a recording surface, and it might be better placed in the `RecordingSurface` type or not?

src/surface.rs
```diff
@@ -165,6 +169,49 @@ impl Surface {
...
+                None
+            } else {
+                Some(Self::from_raw_full(p))
+            }
+        }
+    }
+
+    pub fn unmap_image(&self, image: &Surface) {
+        unsafe {
+            ffi::cairo_surface_unmap_image(self.to_raw_none(), image.to_raw_none())
```
sdroege:
```
The content of the image will be uploaded to the target surface. Afterwards, the image is destroyed.

Using an image surface which wasn't returned by cairo_surface_map_to_image() results in undefined behavior.
```

This needs to check that the surface is of the correct type (needs a surface wrapper type returned by map I guess), and it takes ownership of the image surface.

src/device.rs
sdroege: Recording surface would look like the other specific surface types we have, type-wise. I guess.

src/surface.rs
GuillaumeGomez: I directly used `ImageSurface` type.

sdroege: That's not sufficient. You can only unmap an image surface that you previously got via map here (for the same `self`).

And you need to use `image.to_raw_full()` or something.

sdroege: Also unclear if that's indeed going to be an image surface, always.

src/device.rs
sdroege: There should be a type for the recording surface

GuillaumeGomez: How can we get a recording surface actually?

src/surface.rs
GuillaumeGomez: > Also unclear if that's indeed going to be an image surface, always.

Let's assume it'll be for now.

sdroege: Then you have to make this API unsafe.

sdroege: ... if you don't ensure that it's a surface we got from `map`, I mean.

Also still doing `image.to_raw_none()`. This needs to be `full`.

src/device.rs
sdroege: `cairo_recording_surface_create()`

src/xcb.rs
```diff
@@ -369,77 +369,7 @@ impl Clone for XCBVisualType {
     }
 }
 
-#[derive(Debug)]
-pub struct Device(pub *mut ffi::cairo_device_t);
-
-impl Device {
-    pub fn to_raw_none(&self) -> *mut ffi::cairo_device_t {
```
sdroege: Where did all these functions go?

src/recording_surface.rs
```diff
@@ -0,0 +1,115 @@
...
+impl RecordingSurface {
+    pub fn create<T: Into<Option<Rectangle>>>(
+        content: Content,
+        extends: T,
+    ) -> Option<RecordingSurface> {
+        unsafe {
+            let extends = extends.into();
+            let extends = match extends {
+                Some(c) => c.to_raw_none(),
+                None => 0 as *const _,
```
sdroege: `ptr::null()` is what you want here

src/surface.rs
sdroege: I think what would make sense here is to have `map` return a `struct MappedImageSurface(original_surface, image_surface)`, have it implement `Deref` for the image surface. And on `Drop` it would then call `unmap` and there would be no explicit `unmap` anywhere.

src/xcb.rs
GuillaumeGomez: In the new `Device` type.

src/recording_surface.rs
GuillaumeGomez: Indeed!

src/surface.rs
GuillaumeGomez: Nice idea! I'll implement it this way.

GuillaumeGomez:
Added the `MappedImageSurface` type. :)

GuillaumeGomez:
> `self.image_surface.to_raw_full()`. According to the docs this takes ownership of the argument.

Problem being that there is no `to_raw_full` method implemented anywhere in cairo...

sdroege:
> > `self.image_surface.to_raw_full()`. According to the docs this takes ownership of the argument.
> 
> Problem being that there is no `to_raw_full` method implemented anywhere in cairo...

Then add one here :)

sdroege:
> > > `self.image_surface.to_raw_full()`. According to the docs this takes ownership of the argument.
> > 
> > 
> > Problem being that there is no `to_raw_full` method implemented anywhere in cairo...
> 
> Then add one here :)

Or alternatively to `to_raw_none` and call `ref() on the returned pointer afterwards manually.

.travis.yml
```diff
@@ -54,6 +54,7 @@ script:
   - cargo doc --features "dox,embed-lgpl-docs"
   - cargo test --features "$FEATURES,png,pdf,svg,pdf,xcb"
   - cargo test --no-default-features --features "$FEATURES,png,pdf,svg,pdf,xcb"
+  - cargo test --features "$FEATURES xcb pdf svg png ps use_glib"
```
sdroege: Everywhere else we use commas, here you use spaces now :) And how is this different from `cargo test --features "$FEATURES,png,pdf,svg,pdf,xcb"` (`use_glib` is default feature)

GuillaumeGomez: Woups, local test. Removing it.

cairo-sys-rs/src/lib.rs
```diff
@@ -89,16 +92,14 @@ debug_impl!(cairo_t);
...
-
-#[cfg(any(feature = "xcb", feature = "dox"))]
 #[repr(C)]
 pub struct cairo_device_t(c_void);
-#[cfg(any(feature = "xcb", feature = "dox"))]
 debug_impl!(cairo_device_t);
 
+#[repr(C)]
+#[derive(Copy,Clone,Debug)]
+pub struct cairo_pattern_t(u8);
```
sdroege: Elsewhere we use a tuple struct around `c_void` for this. That's probably a better idea, also to keep it consistent (please check if other things have the same problem).

src/device.rs
```diff
@@ -0,0 +1,347 @@
...
+                    {
+                        ffi::cairo_xcb_device_debug_set_precision(self.to_raw_none(), precision)
+                    }
+                    #[cfg(not(feature = "xcb"))]
+                    {
+                        panic!("you need to enable \"xcb\" feature")
+                    }
+                }
+                _ => {
+                    panic!("invalid device type")
```
sdroege: Maybe here (and elsewhere) you want to print the actual device type in the panic?

src/recording_surface.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright 2015, The Gtk-rs Project Developers.
```
sdroege: It's 2019

src/surface.rs
```diff
@@ -165,6 +171,34 @@ impl Surface {
...
+        }
+    }
+
+    pub fn map_to_image(&self, extents: Option<RectangleInt>) -> Option<MappedImageSurface> {
+        unsafe {
+            let p = match extents {
+                Some(ref e) => ffi::cairo_surface_map_to_image(self.to_raw_none(), e.to_raw_none()),
+                None => ffi::cairo_surface_map_to_image(self.to_raw_none(), 0 as *const _),
+            };
+            if p.is_null() {
```
sdroege: If you read the docs: this function never returns `NULL` but like with many other cairo functions you have to check the status of the returned surface.

sdroege: I.e. it would also return a `Result` and not an `Option`

sdroege:
Otherwise good to go for me

src/recording_surface.rs
GuillaumeGomez: I was so close!

sdroege:
:confetti_ball: let's get it in then :)

GuillaumeGomez:
Thanks for your review!

PR #244, opened 2019-02-15
Implement Display and Debug traits

GuillaumeGomez:
I realised that there were a lot of missing `Debug`/`Display` traits implementations. Going through it.

cc @sdroege @EPashkin

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -6,7 +6,9 @@
 use glib::translate::*;
 use libc::c_int;
 use std::ffi::CString;
+use std::fmt;
 use std::ops;
+use std::slice;
 use ::paths::Path;
 use ::font::{TextExtents, TextCluster, FontExtents, ScaledFont, FontOptions, FontFace, Glyph};
 use ::matrices::{Matrix, MatrixTrait};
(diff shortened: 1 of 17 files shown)


src/context.rs
```diff
@@ -63,6 +63,40 @@ impl Drop for RectangleList {
     }
 }
 
+impl fmt::Debug for RectangleList {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.ptr.is_null() {
+            write!(f, "RectangleList")
```
sdroege: Is this even valid? Judging from other code here it's not so we don't need to handle that here.

src/context.rs
```diff
@@ -63,6 +63,40 @@ impl Drop for RectangleList {
...
+                } else {
+                    format!("[{}]",
+                            slice::from_raw_parts(ptr, len as usize)
+                                .iter()
+                                .map(|s| format!("{:?}", s))
+                                .collect::<Vec<_>>()
+                                .join(", "))
+                };
+
+                write!(f, "RectangleList {{
```
sdroege: Always use the `Formatter::debug_*()` functions for such things, in this case `debug_list()` or even simpler

```rust
f.debug_tuple("RectangleList")
    .field(&*self)  // &* gives you a &[Rectangle]
    .finish()
```

sdroege:
I disagree with all these `Display` impls but keep them if you really want. Just like in glib/etc they don't fulfill the purpose of `Display`.

EPashkin:
@GuillaumeGomez Thanks, 👍 after rebase to master

GuillaumeGomez:
Updated!

sdroege:
Go for it then

GuillaumeGomez:
Yes sir!

Issue #250, opened 2019-03-07
ImageSurface is different then SVG/PDF/PS Surface


jsparber:
I think the API for ImageSurface and the other Surface should be much more similar as they are right now.

cc @meh @vojtechkral @sdroege

vojtechkral:
Well the thing is they are distinct in Cairo itself, see [Image surfaces](https://cairographics.org/manual/cairo-Image-Surfaces.html#cairo-image-surface-create-for-data) and [PDF surfaces](https://cairographics.org/manual/cairo-PDF-Surfaces.html) for example... In particular, image surfaces don't seem to support streaming.

Or do you have other differences in mind?

jsparber:
Also the Svg and PDF are distinct in Cairo, anyhow they should behave all the same in the bindings IMO

sdroege:
@jsparber Maybe you can list the API differences between the different surface types that seem wrong to you, so we can start a discussion about how those specifically could be unified.

ekg:
It would be helpful just to have examples of how to use the svg and pdf surfaces. Right now there are none, and because of the semantic differences with the PNG surface model in cairo, the PNG based example doesn't illuminate how things are bound together for these other types.

vojtechkral:
@ekg There's an example code in the c-tor doc, see [here] (click the little `+` if not displayed right away).

Do you think that's sufficient? How would you suggest the doc/examples be improved?

ekg:
Thank you @vojtechkral, that exactly resolves my problem! I'll see if I can get this to work.

ekg:
I'm now able to render a PDF, but I can't functions like these to write text on it:

```
    let surface = pdf::Writer::new(400.0, 400.0, File::create("test.pdf").unwrap());
    let ctx = Context::new(&surface);
    ctx.scale(120.0, 120.0);
    ctx.select_font_face("Arial", FontSlant::Normal, FontWeight::Normal);
    ctx.set_font_size(18.0);
    ctx.move_to(2.0, 2.0);
    ctx.show_text("hello world");
```

This always yields a blank document, at least under the multiple iterations changing position that I've done. The same code is quite happy from C to draw.

Small examples showing what is needed to get a given capability working would certainly help.

vojtechkral:
@ekg 

> The same code is quite happy from C to draw.

[...]

PR #255, opened 2019-05-09
Expose IO errors in writers, and make them panic-safe

SimonSapin:
This is the `Writer` and `RefWriter` types in the `pdf`, `ps`, and `svg` modules. See individual commit messages for details.

Ideas for version 0.7.0, not included in this PR:

* Replace the `RefWriter` structs with `pub type RefWriter<'w, W> = Writer<&'w mut W>`. (This is technically a breaking change: a dependent crate could implement some trait for both `Writer` and `RefWriter`.)

* Change the return type of `Writer::finish` to include the IO error if any. Maybe `(W, Result<(), io::Error>)`?


src/support.rs
```diff
@@ -21,47 +19,84 @@ pub trait FromRawSurface {
...
+#[derive(Debug)]
+pub struct Writer<S: FromRawSurface + AsRef<Surface>, W: io::Write> {
+    pub surface: S,
+    callback_env: Box<CallbackEnv<W>>,
+}
+
+impl<S: FromRawSurface + AsRef<Surface>, W: io::Write> Writer<S, W> {
+    extern fn write_cb(mut env: *mut c_void, data: *mut c_uchar, length: c_uint) -> cairo_status_t {
+        unsafe {
+            // Safety: the type of `env` would matches `&mut *self.callback_env` in a `Writer` method.
```
GuillaumeGomez: "would matches"?

GuillaumeGomez:
Looks good to me!

> Ideas for version 0.7.0, not included in this PR:
> ...

If you're motivated to implement them, go ahead. :)

GuillaumeGomez:
cc @meh @EPashkin @sdroege

src/support.rs
SimonSapin: Fixed. (Then removed. This was more important when `EnvCallback<W>`

SimonSapin: Fixed. (Then removed. This was more important when `EnvCallback<W>` and `EnvCallback<&mut W>` were both in use.)

SimonSapin:
Implementing them is no problem, but these changes feel small to justify alone an incompatible version bump.

GuillaumeGomez:
Well, next release will be a major one in any case so put your changes at the same time. :)

SimonSapin:
Ah ok, I hadn’t realized there were other breaking changes in the pipeline. I’ve pushed another three commits with the breaking changes proposed above + a rename.

EPashkin:
@SimonSapin Thanks.
Please remove version bump, your publish script do it.

SimonSapin:
Done. How do we tell the script to increment to 0.7.0 instead of 0.6.1?

GuillaumeGomez:
I "tell" him myself when making a release. But like I said, it won't be a minor release in any case.

sdroege:
This is also closely related to https://github.com/gtk-rs/cairo/issues/250

SimonSapin:
Is it? IMO the fix for #250 is more examples in docs, not an API change. (But that’s better discussed there.)

sdroege:
> Is it? IMO the fix for #250 is more examples in docs, not an API change. (But that’s better discussed there.)

It seems to be generally about "the API of these types sucks and should be different" :) I just mentioned it here for completeness as your PR is also about changing the API of these types, but of course in a rather uncontroversial way.

src/support.rs
```diff
@@ -20,48 +18,64 @@ pub trait FromRawSurface {
...
     pub surface: S,
-    writer: Box<W>,
+    callback_env: Box<CallbackEnv<W>>,
 }
 
 impl<S: FromRawSurface + AsRef<Surface>, W: io::Write> Writer<S, W> {
-    extern fn write_cb(writer: *mut c_void, data: *mut c_uchar, length: c_uint) -> cairo_status_t {
-        let mut writer: Box<W> = unsafe { Box::from_raw(writer as *mut _) };
+    extern fn write_cb(env: *mut c_void, data: *mut c_uchar, length: c_uint) -> cairo_status_t {
+        let env: &mut CallbackEnv<W> = unsafe { &mut *(env as *mut CallbackEnv<W>) };
```
sdroege: Should we check here if an error was already set before but for some reason we were called again, and in that case immediately error out?

src/support.rs
```diff
@@ -20,48 +18,64 @@ pub trait FromRawSurface {
...
     }
 
     pub fn new(constructor: Constructor, width: f64, height: f64, writer: W) -> Writer<S, W> {
-        let mut writer = Box::new(writer);
-        let writer_ptr = unsafe { mem::transmute(&mut *writer) };
+        let mut callback_env = Box::new(CallbackEnv {
+            writer,
+            error: None,
+        });
+        let env_ptr = &mut *callback_env as *mut CallbackEnv<W> as *mut c_void;
```
sdroege: Yay, getting rid of an unneeded transmute :fireworks:

src/support.rs
```diff
@@ -20,48 +18,64 @@ pub trait FromRawSurface {
...
 
-    pub fn finish(self) -> W {
-        let surface = self.surface;
-        surface.as_ref().finish();
-        drop(surface);
+    pub fn io_error(&self) -> Option<&io::Error> { self.callback_env.error.as_ref() }
+    pub fn take_io_error(&mut self) -> Option<io::Error> { self.callback_env.error.take() }
 
-        *self.writer
+    pub fn into_inner(self) -> Result<W, (io::Error, W)> {
```
sdroege: It would be nice if we could somehow force the user to call finish/into_inner at the end instead of letting the type be dropped silently. It's most likely a bug if it's not done, or am I missing something?

Not sure how we could do that nicely though.

sdroege:
Generally looks good to me, thanks :)

src/support.rs
SimonSapin: Good point!

I think both behaviors are reasonable, I don’t have a strong preference. Not attempting to write again after there has been an error makes sense. I wonder if cairo does the same after the callback returns `CAIRO_STATUS_WRITE_ERROR`

src/support.rs
SimonSapin: The type theory term for this is linear types (use exactly once). In Rust today, `!Copy` types are called affine (use at most once, where borrows do not count as a “use”). https://gankro.github.io/blah/linear-rust/ explains how it would be tricky to add linear types to the language.

sdroege: I was thinking of putting a panic into the `Drop` impl (ugly, at runtime) or somehow doing something with `warn_unused` but not sure how that could be nicely done here

SimonSapin:
Closing in favor of https://github.com/gtk-rs/cairo/pull/260.

PR #257, opened 2019-05-12
Add a safe API for user data

SimonSapin:


SimonSapin:
Should doc-comments go into https://github.com/gtk-rs/lgpl-docs? Are docs written specifically for Rust bindings also under the LGPL?

GuillaumeGomez:
If the doc comments come from somewhere under LGPL, then LGPL, otherwise it's MIT and you can leave them where they are.

src/image_surface.rs
```diff
@@ -48,23 +49,26 @@ impl ImageSurface {
...
-            }
+        let result = unsafe {
+            ImageSurface::from_raw_full(
+                ffi::cairo_image_surface_create_for_data(ptr, format.into(), width, height, stride)
+            )
+        };
+        if let Ok(surface) = &result {
+            static IMAGE_SURFACE_DATA: crate::UserDataKey<Box<dyn AsMut<[u8]>>> =
+                crate::UserDataKey::new();
+            surface.set_user_data(&IMAGE_SURFACE_DATA, Rc::new(data))
```
sdroege: Why does this have to be in an `Rc`?

sdroege: Oh I see, because of the getter

sdroege:
:+1:

GuillaumeGomez:
Thanks @SimonSapin (especially for the documentation!)!

Issue #258, opened 2019-05-12
Rework type hierarchies?


SimonSapin:
Since breaking changes are on the table already (https://github.com/gtk-rs/cairo/pull/255#issuecomment-491065296) I’m considering some more, but I’d like some opinions before writing code.

# Patterns

cairo-rs 0.6.0 has a `Pattern` enum where each variant contains a pointer-wrapping struct for the corresponding kind of pattern. Both the enums and each struct implement a `PatternTrait` trait that has methods in common to all patterns. Two of the structs implement a `Gradient` trait.

For bindings in object-oriented languages, https://www.cairographics.org/manual/bindings-patterns.html reccomends classes for each pattern kind and inheritance. The closest in today’s Rust is having structs whose only (or first, with `#[repr(C)]`) field is the parent type, that implement `Deref<Target=ParentType>` (and maybe `AsRef<AncestorType>` for all recursive ancestors, I’m not sure it’s needed given auto-deref). cairo-rs 0.6.0 already does this with surfaces.

This way, the `Pattern` and `Gradient` structs could have inherent methods that are usable on “sub-types” through auto-deref, without the need for traits.

Additionally, there could be methods similar to the standard library’s [`downcast`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.downcast) to go from less specific to more specific types.

# PDF, PS, and SVG surfaces

https://github.com/gtk-rs/cairo/issues/250 has some complaints about the API for these surface types. While I disagree with some of the points there (there’s a reason for differences with `ImageSurface`, they are already closer to idiomatic Rust than bare C FFI, …) I do agree that the current API is a bit weird:

There are public modules named `pdf`, `ps`, and `svg` (whereas all other modules except `prelude` are private with contents reexported from the crate root) and contain identical types named `File`, `Writer`, and `RefWriter`.

I think we can have one Rust type for each of these surface kinds, and no public module for them. For example `PdfSurface` with `create` and `create_for_stream` constructors.

(BTW I’m not sure we can safely keep `RefWriter`’s functionality, for the same reason as https://github.com/gtk-rs/cairo/issues/228. (Surface ref-counting make it insufficient to have a wrapper type with a lifetime parameter.) It may still be worth having an `unsafe fn` constructor for it.)

# Remaining public traits

* `XCBSurface` is a trait implemented for `Surface`, but https://github.com/gtk-rs/cairo/pull/102 does not explain why. Should it be a type similar to `ImageSurface` instead?

[...]

PR #260, opened 2019-05-20
Rework type hierarchies

SimonSapin:
This is a series of breaking API changes that implement the proposals from https://github.com/gtk-rs/cairo/issues/258.

The following methods are added on `Surface`, although they are only relevant to surfaces created with `{Pdf,Ps,Svg}Surface::for_{raw_,}stream`. (They return `None` or `Ok(())` in other cases.)

```rust
pub fn take_output_stream(&self) -> Option<Box<dyn Any>> {…}
pub fn take_io_error(&self) -> Result<(), io::Error> {…}
pub fn borrow_output_stream<'surface>(&'surface self) -> Option<RefMut<'surface, dyn Any>> {…}
pub fn borrow_io_error<'surface>(&'surface self) -> Result<(), RefMut<'surface, io::Error>> {…}
```


src/stream.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /// Remove and return the output stream, if any.
+    ///
+    /// This is relevant for surfaces created for example with [`PdfSurface::for_stream`].
+    /// Consider calling [`Surface::finish`] first,
+    /// to ensure that all writes to the stream are done.
+    ///
+    /// Use [`Box::downcast`] to recover the concrete type.
+    pub fn take_output_stream(&self) -> Option<Box<dyn Any>> {
```
SimonSapin: I’ve put these public methods on `Surface`, but they could instead be on `PdfSurface`, `PsSurface`, and `SvgSurface`.

src/patterns.rs
```diff
@@ -20,187 +22,151 @@ use ffi::{
...
-    unsafe fn from_raw_full(pointer: *mut cairo_pattern_t) -> Self::PatternType;
-
-    unsafe fn from_raw_none(pointer: *mut cairo_pattern_t) -> Self::PatternType {
-        ffi::cairo_pattern_reference(pointer);
-        Self::from_raw_full(pointer)
+impl Clone for Pattern {
+    fn clone(&self) -> Self {
+        Pattern {
+            pointer: unsafe {
+                ffi::cairo_pattern_reference(self.pointer)
```
GuillaumeGomez: Shouldn't you check if the returned pointer is not NULL in here?

src/pdf.rs
```diff
@@ -387,10 +261,11 @@ mod test {
 
         let custom_writer = CustomWriter(0);
 
-        let surface = Writer::new(20., 20., custom_writer);
+        let surface = PdfSurface::for_stream(20., 20., custom_writer);
         surface.set_size(100., 100.);
         draw(&surface);
-        let custom_writer = surface.finish();
+        surface.finish();
+        let custom_writer = surface.take_output_stream().unwrap().downcast::<CustomWriter>().unwrap();
```
GuillaumeGomez: Minor: Please use `expect` instead of `unwrap`, makes errors easier to spot when they occur (even more in tests!).

src/patterns.rs
SimonSapin: This function returns the same pointer that it received, which must be non-NULL. https://www.cairographics.org/manual/cairo-cairo-pattern-t.html#cairo-pattern-reference

src/ps.rs
```diff
@@ -310,46 +180,47 @@ mod test {
     fn draw_in_buffer() -> Vec<u8> {
         let buffer: Vec<u8> = vec![];
 
-        let surface = Writer::new(100., 100., buffer);
+        let surface = PsSurface::for_stream(100., 100., buffer);
         draw(&surface);
-        surface.finish()
+        surface.finish();
+        *surface.take_output_stream().unwrap().downcast().unwrap()
```
GuillaumeGomez: `expect`s please. :)

src/stream.rs
sdroege: It would seem nicer to not pollute the normal `Surface` API with that but have it only on the types where it can actually exist, just like the data related functions on `ImageSurface`.

What was your reasoning for having it on the main `Surface` type? Maybe I'm missing an advantage of doing so :)

src/image_surface_png.rs
```diff
@@ -14,61 +16,86 @@ use ImageSurface;
...
-        },
+        }
     }.into()
 }
 
 struct WriteEnv<'a, W: 'a + Write> {
     writer: &'a mut W,
-    error: Option<Error>,
+    io_error: Option<Error>,
+    unwind_payload: Option<Box<dyn Any + Send + 'static>>,
```
sdroege: Can we have unwind and error? Maybe makes sense to use `Either` here to make it more explicit that only one of the cases can exist, but I'm fine with this here too :)

src/stream.rs
```diff
@@ -138,8 +145,11 @@ static STREAM_CALLBACK_ENVIRONMENT: UserDataKey<UnsafeCell<CallbackEnvironment>>
 struct CallbackEnvironment {
     stream: Option<Box<dyn Any>>,
     io_error: Option<io::Error>,
+    unwind_payload: Option<Box<dyn Any + Send + 'static>>,
 }
 
+// Safety: unwinding into C is undefined behavior (https://github.com/rust-lang/rust/issues/58794)
+// so code outside of the `catch_unwind` call must never panic.
```
sdroege: Isn't this actually defined behaviour now (panic) since one of the last few releases? But your solution is nicer of course.

We depend on this being defined behaviour in the other crates. Started to do so in 1.24, then it was backed out again for 1.25, and we were told that it was going to be stabilized again in some 1.3x release I can't remember right now. All this back and forth about this behaviour is really annoying.

src/image_surface.rs
```diff
@@ -15,25 +16,29 @@ use ::enums::{
...
+use surface::Surface;
 use Status;
 use std::fmt;
 
 #[derive(Debug)]
 pub struct ImageSurface(Surface);
 
-impl ImageSurface {
-    pub fn from(surface: Surface) -> Result<ImageSurface, Surface> {
+impl TryFrom<Surface> for ImageSurface {
```
sdroege: Should make sure in the release notes that we communicate that we only support 1.34+ now.

src/image_surface_png.rs
```diff
@@ -14,61 +16,86 @@ use ImageSurface;
...
+            write_env.unwind_payload = Some(payload);
             Status::WriteError
-        },
+        }
     }.into()
 }
 
 
 impl ImageSurface {
     pub fn create_from_png<R: Read>(stream: &mut R) -> Result<ImageSurface, IoError> {
```
sdroege: How is this safe, what ensures that the `&mut R` stays alive for as long as the `ImageSurface`?

src/patterns.rs
```diff
@@ -20,187 +22,151 @@ use ffi::{
...
-#[derive(Debug, Clone)]
-pub enum Pattern {
-    SolidPattern(SolidPattern),
-    SurfacePattern(SurfacePattern),
-    LinearGradient(LinearGradient),
-    RadialGradient(RadialGradient),
-    Mesh(Mesh),
+#[derive(Debug)]
+pub struct Pattern {
+    pointer: *mut cairo_pattern_t,
```
sdroege: `ptr::NonNull`?

src/stream.rs
```diff
@@ -0,0 +1,229 @@
...
+        ///
+        /// The value that `stream` points to must live at least until the underlying `cairo_surface_t`
+        /// (which maybe be longer then the Rust `PdfSurface` wrapper, because of reference-counting),
+        /// or until the output stream is removed from the surface with [`Surface::take_output_stream`].
+        ///
+        /// Since the former is hard to track for sure, the latter is strongly recommended.
+        /// The concrete type behind the `Box<dyn Any>` value returned by `take_output_stream`
+        /// is private, so you won’t be able to downcast it.
+        /// But removing it anyway ensures that later writes do no go through a dangling pointer.
+        pub unsafe fn for_raw_stream<W: io::Write + 'static>(width: f64, height: f64, stream: *mut W) -> Self {
```
sdroege: Should we also remove the `'static` here to make it more generic and unsafe? :)

src/stream.rs
```diff
@@ -0,0 +1,229 @@
...
+
+impl<W: io::Write> io::Write for RawStream<W> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> { unsafe { (*self.0).write(buf) }}
+    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> { unsafe { (*self.0).write_all(buf) }}
+    fn flush(&mut self) -> io::Result<()> { unsafe { (*self.0).flush() } }
+}
+
+trait AnyExt {
+    /// Any::downcast_mut, but YOLO
+    unsafe fn downcast_mut_unchecked<T>(&mut self) -> &mut T {
```
sdroege: Does it make any measurable performance difference compared to the safe version?

src/patterns.rs
sdroege: Should also create an issue to transform any other types to that instead of plain pointers, unless you want to do this as part of this PR too.

sdroege:
Thanks, looks generally good to me :)

src/image_surface_png.rs
SimonSapin: With the current behavior it’s possible to have both. However we may want to change the behavior to not attempt another write/read if a previous one returned an error or panicked.

src/stream.rs
SimonSapin: As the linked issue https://github.com/rust-lang/rust/issues/58794 explains, abort-on-unwind-through-`extern fn` was landed then reverted. Twice. But even if it lands again, isn’t propagating the panic preferable to aborting the process?

src/patterns.rs
SimonSapin: I think that would work, here and for many (every?) other structs that wrap cairo pointers. But this PR is already large and using `NonNull` internally doesn’t affect the public API, so I’m happy to leave for later or someone else.

src/image_surface_png.rs
sdroege: Seems reasonable to have all future read/writes fail

src/stream.rs
SimonSapin: We can’t easily. That bound is required to keep even `*mut W` in a `Box<dyn Any>`.

src/stream.rs
sdroege: In this specific case here we can do that, in general we unfortunately can't (there might be no well known place from which we would propagate the panic further) so aborting the process is the best we can do.

Not sure if we want to, yet another time, revert back to adding guards to the other crates to exit the process at the FFI boundary. It's getting boring and causes a lot of churn, and the undefined behaviour right now seems to be to abort the process...

Not complaining to you personally here btw :) Thanks for providing the link to the latest version of the adventure.

src/stream.rs
sdroege: Makes sense, thanks

src/patterns.rs
sdroege: Works for me :)

sdroege: https://github.com/gtk-rs/cairo/issues/261

src/stream.rs
SimonSapin: I haven’t measured it, but I’m relatively confident about this particular aspect of the unsafety of the implementation. See the comment at the call site: https://github.com/gtk-rs/cairo/pull/260/files/cf2f6925ffe09f45429afd1940731b57454c6d49#diff-2763e33724cc9b239c750ae973480953R168

I did encounter segfauts initially when this method was a free function taking a `&mut dyn Any` parameter. I would pass it `&mut Box<dyn Any>` expecting auto-deref, but the unsizing coercion of `Box<dyn Any>` to `dyn Any` (adding another layer of trait object indirection) took precedence. So I would end up with a reference at the wrong address, even though `debug_assert!(stream.is::<W>());` passed! Making this a method in an extension trait fixes this because unsizing coercion desn’t happen for method receivers like it does for call arguments.

sdroege: Ok, it *seems* fine to me too so let's keep it then :) Thanks for the details

src/image_surface_png.rs
SimonSapin: Done.

SimonSapin:
I’ve realized that my use of `UnsafeCell` was unsound. Even though it feels that you would have to actively go out of your way to cause trouble, it is possible in safe code to make a type that implements `io::Write` and owns a reference-counted clone of the `Surface`, then for example use `take_output_stream` to end up with multiple `&mut` references to itself.

I’ve switched to a `RefCell` and using `try_borrow_mut` in the callback to avoid `borrow_mut`’s panic, and a `Cell<bool>` to keep track of failed borrows. (We do handle panics, but need mutable access to the environment to save the payload!)

This enables safely borrowing the stream or I/O error without removing them (by returning a `std::cell::RefMut` after using `RefMut::map`), so I’ve replaced the boolean methods with that. The new APIs for all surfaces are now:

```rust
pub fn take_output_stream(&self) -> Option<Box<dyn Any>> {…}
pub fn take_io_error(&self) -> Result<(), io::Error> {…}
pub fn borrow_output_stream<'surface>(&'surface self) -> Option<RefMut<'surface, dyn Any>> {…}
pub fn borrow_io_error<'surface>(&'surface self) -> Result<(), RefMut<'surface, io::Error>> {…}

**Comment by sdroege** [comment]
Nice!

File: src/stream.rs (hunk 6 above)
Comment (SimonSapin) [reply src/stream.rs]: > there might be no well known place from which we would propagate the panic further

Indeed, https://github.com/gtk-rs/cairo/issues/262 is such a case.

The PDF / PS / SVG output stream case is sort of in-between. And we’re not alone with these questions, the end of https://www.cairographics.org/manual/bindings-streams.html discusses it as well.

**Comment by vojtechkral** [comment]
I'll try to have a look at some point today...

**Comment by SimonSapin** [comment]
What’s the way to deal with CI failure because of code in other repositories? I have a hard time even reproducing the failure.

File: src/pdf.rs [hunk 12]
```diff
@@ -358,10 +228,14 @@ mod test {
...
-        let surface = RefWriter::new(100., 100., &mut file);
+        let surface = unsafe {
+            PdfSurface::for_raw_stream(100., 100., &mut file)
+        };
 
         draw(&surface);
         surface.finish();
+        surface.take_output_stream();
+        drop(file);
     }
```
vojtechkral: Since we discussed with @sdroege that writing to a stream after `finish()` would result in an error, it would be nice to perhaps test for that as well, to see if it's handled gracefully and stuff.

sdroege:
> What’s the way to deal with CI failure because of code in other repositories? I have a hard time even reproducing the failure.

For the docs related failure, that's @GuillaumeGomez special area of expertise :)

The failure in gdk needs a PR to be submitted here: https://github.com/gtk-rs/gdk
And we'll simply ignore that CI failure here, merge the cairo PR, then merge the gdk PR once its CI succeeded. Unfortunately there's no clean way implemented right now to make the CI happy all the way.

Btw, please also remove the unneeded `mut` in image_surface.rs :)

GuillaumeGomez:
Weird. The `rustdoc-stripper` might not work perfectly in macros...

src/pdf.rs
vojtechkral: Err, actually it looks like this isn't implemented, looks like `write_callback()` does nothing if the stream is taken. This is what I originally suggested, but I think @sdroege is right in that this should instead be an error. Also, seeing the code in tests, it occurs to me that it might make sense to call `finish()` as part of `take_output_stream()`. Does it make sense to call `take_output_stream()` without calling `finish()` right before that? I can't think of a situation where that would make sense...

vojtechkral: Ok, sorry for the fragmented comments, but it seems like cairo [already handles](https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-finish) this by setting an error state if drawing is done after `finish()`, so it would seem to me the right choice would be to make `take_output_stream()` imply `finish()` which would in turn result in error status in case of drawing operations comin in afterwards.

vojtechkral:
Might also be a good idea to test that panicking in streams works, such as:

```Rust
#[test]
#[should_panic]
fn panic_writer() {
    struct PanicWriter;

    impl io::Write for PanicWriter {
        fn write(&mut self, buf: &[u8]) -> io::Result<usize> { panic!("panic in writer"); }
        fn flush(&mut self) -> io::Result<()> { panic!("panic in writer"); }
    }

    let surface = PdfSurface::for_stream(20., 20., PanicWriter);
    surface.set_size(100., 100.);
    draw(&surface);
    surface.finish();
    surface.take_output_stream();
}
```

sdroege:
What's the status here? Is everybody happy with the changes, are all review comments handled? :)

vojtechkral:
> What's the status here? Is everybody happy with the changes, are all review comments handled? :)

I think mine aren't. I have two suggestions:

- Call `finish()` as part of `take_output_stream()` (if you guys agree)
- Add the panic test above (ditto)

SimonSapin:
I’ve added some tests.

As the last commit (48ee818b24aae56e79c76443b8c50cc079a7c24b) shows, it’s easy to forget to check for IO errors and accidentally ignore them.

Calling `finish` in `take_output_stream` would exacerbate this, since `finish` can (and typically does) cause more writes to happen and these writes can return an IO error. So it might be better to have a single method that combines `finish`, checking for error, and returning the stream. However the signature can quickly become unwieldy.

```rust
fn finish_output_stream(&self) -> (Option<Box<dyn Any>>, Result<(), io::Error>)
```

However a tuple doesn’t benefit from `#[must_use]`, so a top-level `Result` might be preferable. But a user might want recover the stream even in the error case.

```rust
fn finish_output_stream(&self) -> Result<Option<Box<dyn Any>>, (io::Error, Option<Box<dyn Any>>)>
```

Maybe we culd simply a bit by removing the `Option` layer and returning `Box<dyn Any>` of a private zero-size type instead of `None` (when the output stream was already taken, or there wasn’t one in the first place). This would make `downcast` always return an error in those cases (since users do not have access to this private type). This makes user code less noisy (with potentially fewer `unwrap` calls) but maybe it becomes less easy to follow everything that goes on.

```rust
fn finish_output_stream(&self) -> Result<Box<dyn Any>, (io::Error, Box<dyn Any>)>
```

A `Result` value can be used with the `?` operator, but perhaps not easily if the error type is a tuple that doesn’t implement any conversion. So we could add a dedicated error type:

```rust
impl Surface {
    fn finish_output_stream(&self) -> Result<Box<dyn Any>, ErroredStream> {…}
}
pub struct ErroredStream {
    pub error: io::Error,
    pub stream: Box<dyn Any>,
}
impl From<ErroredStream> for io::Error {
    fn from(e: ErroredStream) -> Self {
        e.error
    }
}
```

… but yet another type adds “weight” to the API and doesn’t feel great.


Any opinion?

GuillaumeGomez:
@SimonSapin: I have a preference for the last one. Yet another type but so much easier to read than the previously suggested...

sdroege:
I agree, the last one seems cleanest. And it probably makes sense to not provide a non-finishing `take` function for the stream but to ensure that the surface is first finished (i.e. only have the new function you propose here). It does not seem nice or useful to take away the stream before the surface is finished.

vojtechkral:
@SimonSapin Thanks for thinking through the various options here. FWIW I'd also say the last one is preferable.

Nit: Perhaps the error conversion impl could be generalized like:

```Rust
impl From<ErroredStream> for T: From<io::Error> {
    fn from(e: ErroredStream) -> Self {
        Self::from(e.error)
    }
}
```

(Hope I'm not getting into bikeshedding teritory too much.)

SimonSapin:
Eh, sorry I wasn’t clear. I didn’t mean this as a closed set of signatures to choose from. Rather, there’s a series of trade-offs that don’t have an obvious answer to me. For example:

* Is an `Option` useful to separately signal cases where the surface doesn’t have an output stream (or where it was already taken), or is `Box<dyn Any>` of a dummy private concrete type a good idea and not too hard to understand?
  - Or even should these APIs panic it those cases?
  - Should the answer be the same for the two cases?
* Is it worth having the `Result` at the "top-level" even if this makes the signature more complex? (With the stream type in two places)
  - If so, does a “stream + IO error” struct carry its weight compare to a tuple?
  - Or should we have neither because no-one even care about recovering the stream in the error case?

----

@vojtechkral I think that impl coherence rules would not allow this, because neither `T` nor `From` is crate-local. And even if it did, this kind of overly-generic API can become hard to use as more turbofishes are required because type inference doesn’t have enough context to determine an "intermediate" type parameter.

sdroege:
> Is an `Option` useful to separately signal cases where the surface doesn’t have an output stream (or where it was already taken), or is `Box<dyn Any>` of a dummy private concrete type a good idea and not too hard to understand?
> 
>     * Or even should these APIs panic it those cases?
> 
>     * Should the answer be the same for the two cases?

My vote here would be to panic. If it's called on a surface that the user did not create with a stream, then there's something very wrong in the user code.

> Is it worth having the `Result` at the "top-level" even if this makes the signature more complex? (With the stream type in two places)
> 
>     * If so, does a “stream + IO error” struct carry its weight compare to a tuple?
> 
>     * Or should we have neither because no-one even care about recovering the stream in the error case?

IMHO the struct with stream+error in a Result gives the nicest API here

GuillaumeGomez:
> My vote here would be to panic. If it's called on a surface that the user did not create with a stream, then there's something very wrong in the user code.

If it potentially comes from user code, then I'd prefer avoid panics if possible. But not sure if this is better in here...

sdroege:
> If it potentially comes from user code, then I'd prefer avoid panics if possible. But not sure if this is better in here...

What do you expect the user to do, other than panicking, with the `None` that they expect to be `Some(stream)` in their code? :)

GuillaumeGomez:
Basically yes. :laughing:

sdroege:
Thanks, lgtm :)

SimonSapin:
Alright. The signature is now:

```rust
pub fn finish_output_stream(&self) -> Result<Box<dyn Any>, StreamWithError>
```

It panics when called more than once, or if there was no output stream to begin with.

I’ve removed the `borrow_*` methods. While it’s cool that they can safely be implemented, they’re not useful without also calling `Surface::finish` at which point the surface pretty much can’t be used anymore so one might as well call `finish_output_stream` instead.

sdroege:
@vojtechkral @GuillaumeGomez @gekola @meh Does it look good to all of you too? IMHO we're done here now :)

EPashkin:
LGFM, but doc-check better fixed by updating lgpl-docs or adding ` src/stream.rs` etc. to `IGNORES` in `build.rs`

vojtechkral:
lgtm, @SimonSapin thanks for doing all this

GuillaumeGomez:
As promised to @SimonSapin, he did his part. I'll just check what's missing on the other repositories then merge this PR. Thanks a lot @SimonSapin !

sdroege:
Indeed, thanks a lot for cleaning this up @SimonSapin :)

GuillaumeGomez:
It'll require to update gdk. I'll do it shortly. Thanks a lot @SimonSapin !

PR #264, opened 2019-06-11
Set PKG_CONFIG_PATH in macOS build script

cmyr:
This is a... speculative patch. It seems to _work_, but I don't feel great about it. I'm out of my depth here so I'm not really sure what the best approach would be, but this should at least get the crate to build for the 95+% of macOS users with homebrew.

I am *very* open to suggestions for how to do this more cleanly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,6 @@ addons:
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install gtk+3 cairo atk; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-    export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig:/usr/local/opt/libffi/lib/pkgconfig;
-    fi
 script:
   - rustc --version
   - cargo doc --features "dox,embed-lgpl-docs"
--- a/cairo-sys-rs/build.rs
+++ b/cairo-sys-rs/build.rs
@@ -14,6 +14,10 @@ fn main() {
             process::exit(1);
         }
     } else {
+        #[cfg(target_os = "macos")]
+        {
+            macos_maybe_find_lib_ffi();
+        }
         if let Err(s) = find("cairo", &["cairo"]) {
             let _ = writeln!(io::stderr(), "{}", s);
             process::exit(1);
@@ -68,5 +72,46 @@ fn find(package_name: &str, shared_libs: &[&str]) -> Result<(), Error> {
             Ok(())
         }
         Err(err) => Err(err),
+    }
+}
+
+// HACK: if libffi is installed via homebrew it doesn't get linked correctly.
+// this adds the homebrew libffi to the pkg-config-path.
+#[cfg(target_os = "macos")]
+fn macos_maybe_find_lib_ffi() {
+    use std::path::{Path, PathBuf};
+    use std::process::Command;
+
+    const PKG_CONFIG_PATH: &str = "PKG_CONFIG_PATH";
+    const BREW_LIBFFI_REL_PATH: &str = "opt/libffi/lib/pkgconfig";
+
+    // check for homebrew
+    if !Command::new("brew")
+        .arg("-v")
+        .output()
+        .map(|o| o.status.success())
+        .unwrap_or(false)
+    {
+        return;
+    }
+
+    if let Some(prefix) = Command::new("brew")
+        .arg("--prefix")
+        .output()
+        .ok()
+        .and_then(|o| String::from_utf8(o.stdout).ok())
+        .map(|s| PathBuf::from(s.trim()))
+    {
+        let libffi_path = prefix.join(BREW_LIBFFI_REL_PATH);
+        let cfg_path = env::var(PKG_CONFIG_PATH)
+            .ok()
+            .and_then(|var| env::join_paths(&[Path::new(&var), libffi_path.as_path()]).ok())
+            .unwrap_or(libffi_path.into());
+        env::set_var(PKG_CONFIG_PATH, cfg_path);
+    } else {
+        eprintln!(
+            "brew --prefix returned non-utf8 output, \
+             so you probably know what you're doing."
+        );
     }
 }


cairo-sys-rs/build.rs
```diff
@@ -70,3 +72,19 @@ fn find(package_name: &str, shared_libs: &[&str]) -> Result<(), Error> {
...
+fn macos_maybe_find_lib_ffi() {
+    #[cfg(target_os = "macos")]
+    {
+        let brew_prefix = process::Command::new("brew")
+            .arg("--prefix")
+            .output()
+            .map(|o| PathBuf::from(String::from_utf8_lossy(&o.stdout).trim()))
+            .unwrap_or(PathBuf::from("/usr/local"));
+        let libffi_path = brew_prefix.join("opt/libffi/lib/pkgconfig");
+        env::set_var("PKG_CONFIG_PATH", libffi_path);
```
GuillaumeGomez: If the `PKG_CONFIG_PATH` variable is already set, you're completely overwriting it... Not very ideal. :-/

cairo-sys-rs/build.rs
```diff
@@ -14,6 +15,7 @@ fn main() {
             process::exit(1);
         }
     } else {
+        macos_maybe_find_lib_ffi();
```
GuillaumeGomez: It'd be better to call this function only on macOS instead of making the condition inside the function.

cairo-sys-rs/build.rs
cmyr: should I infer from this review that you would consider taking this patch if it were less bad?

GuillaumeGomez:
I'm clearly not a big fan of this solution. However, I have no idea how to solve this issue otherwise...

What do you think @sdroege @EPashkin ?

cairo-sys-rs/build.rs
cmyr: that was how I'd written it originally but I thought it might be better to not clutter up the call-site with the cfg annotation. happy to make this change, though.

cmyr:
we're on the same page re: not liking this approach, it's just the only one I can think of. 😒

cairo-sys-rs/build.rs
GuillaumeGomez: Honestly yes. It's been an annoying issue for far too long already. I never did it myself because I felt like it seemed way too much but I wasn't able to come up with a solution in the meantime, so...

EPashkin:
I 👍 for this way, if it not corrupt current PKG_CONFIG_PATH

cmyr:
okay cleaned this up a bit, let me know if you have any further suggestions.

cairo-sys-rs/build.rs
```diff
@@ -70,3 +75,40 @@ fn find(package_name: &str, shared_libs: &[&str]) -> Result<(), Error> {
         Err(err) => Err(err),
     }
 }
+
+// HACK: if libffi is installed via homebrew it doesn't get linked correctly.
+// this adds the homebrew libffi to the pkg-config-path.
+fn macos_maybe_find_lib_ffi() {
```
GuillaumeGomez: Please add the condition here too.

cairo-sys-rs/build.rs
```diff
@@ -4,7 +4,8 @@ use pkg_config::{Config, Error};
 use std::env;
 use std::io;
 use std::io::prelude::*;
-use std::process;
+use std::path::{Path, PathBuf};
```
GuillaumeGomez: In here: two possibilities:

 1. you move the import into the function so no need to put conditional compilation on them
 2. you put conditional compilation on them

:)

GuillaumeGomez:
Also, can you remove the `PKG_CONFIG_PATH` update in the `.travis.yml` file please? It'll allow us to check if it's working as expected.

cmyr:
alright, made those changes, let's see how CI does.

.travis.yml
```diff
@@ -47,9 +47,6 @@ addons:
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install gtk+3 cairo atk; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
```
GuillaumeGomez: I'm so happy to see this disappear. T_T

cairo-sys-rs/build.rs
```diff
@@ -70,3 +74,43 @@ fn find(package_name: &str, shared_libs: &[&str]) -> Result<(), Error> {
...
     }
 }
+
+// HACK: if libffi is installed via homebrew it doesn't get linked correctly.
+// this adds the homebrew libffi to the pkg-config-path.
+#[cfg(target_os = "macos")]
+fn macos_maybe_find_lib_ffi() {
+    use std::path::{Path, PathBuf};
+    use std::process::Command;
+    const PKG_CONFIG_PATH: &str = "PKG_CONFIG_PATH";
```
GuillaumeGomez: Small nit: Please add an empty line after the imports, it makes code easier to read.

GuillaumeGomez:
It failed on macOS:

```
error: failed to run custom build command for `gobject-sys v0.8.0 (https://github.com/gtk-rs/sys#c45adc74)`
process didn't exit successfully: `/Users/travis/build/gtk-rs/cairo/target/debug/build/gobject-sys-61d058cef7f239df/build-script-build` (exit code: 1)
--- stderr
`"pkg-config" "--libs" "--cflags" "gobject-2.0 >= 2.42"` did not exit successfully: exit code: 1
--- stderr
Package libffi was not found in the pkg-config search path.
Perhaps you should add the directory containing `libffi.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libffi', required by 'gobject-2.0', not found
warning: build failed, waiting for other jobs to finish...
```

cmyr:
hmm, was maybe a bit ambitious; this is working for `cairo-sys-rs` but not for `cairo`. Not sure if anyone here has any ideas, I'll poke around a bit; would obviously prefer to not have this in _two_ build files 😒

sdroege:
This is a bug in brew and the workaround like this will break other setups where explicitly a different version of GTK and its dependencies is selected via the `PKG_CONFIG_PATH` variable. Specifically it breaks my use-case.

The problem here is that somehow `gobject-2.0.pc` is in `PKG_CONFIG_PATH` (or the default path of pkg-config) but not all of its dependencies (`libffi` in this case). Did someone ever report this to brew or are we just hacking around things here? :)

That said, it's glib that depends on libffi so you probably want to add some kind of workaround to the `glib-sys` build.rs and not cairo. An acceptable workaround would seem to check the output of `pkg-config` on macOS and if `gobject-2.0` is found but its dependency `libffi` not, then this workaround can be applied. But report this issue to brew as a very first step.

sdroege:
Also note that the other `build.rs` are autogenerated by the `gir` tool, so having the workaround in multiple files is fine. Alternatively we might want to introduce some gtk-rs-pkg-config-workarounds crate for all these things. There are other workarounds in there already, specifically for Windows being a big mess.

cmyr:
so looking at `brew info libffi`, I get:

```
libffi is keg-only, which means it was not symlinked into /usr/local,
because some formulae require a newer version of libffi.

For compilers to find libffi you may need to set:
  export LDFLAGS="-L/usr/local/opt/libffi/lib"

For pkg-config to find libffi you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/libffi/lib/pkgconfig"
```

So it isn't clear to me that this is a bug in homebrew, per se. (Also: I will reemphasis that I don't know what I'm talking about.)

playing around, I can build `glib-sys` on its own without any problem; and `cairo-sys-rs` fails to build regardless of whether I pass `--features=use_glib`.

I'm not sure what the next step should be, so if anyone has any ideas, please advise!

sdroege:
> I'm not sure what the next step should be, so if anyone has any ideas, please advise!

So I guess that it does not work out of the box is by design and you have to set `PKG_CONFIG_PATH` yourself accordingly like what you've quoted there. That works, or doesn't it?

I disagree and would consider this a bug: if they put `gobject-2.0.pc` into the default `PKG_CONFIG_PATH`, then they must also put `libffi.pc` in there or otherwise the former will simply not work. And it must be from the same version also.

cmyr:
yes, running `PKG_CONFIG_PATH="/usr/local/opt/libffi/lib/pkgconfig" cargo build` works. 

There is an issue open in homebrew (https://github.com/Homebrew/homebrew-core/issues/40179) but it doesn't seem especially encouraging.

GuillaumeGomez:
I wonder if the `.set_var(...)` isn't applied only on cargo specifically?

sdroege:
> I wonder if the `.set_var(...)` isn't applied only on cargo specifically?

What do you mean? Which `set_var`?

GuillaumeGomez:
This [one](https://github.com/gtk-rs/cairo/pull/264/files#diff-8b48b2b6c1e1fa25f5d88692710e80fbR110).

sdroege:
> This [one](https://github.com/gtk-rs/cairo/pull/264/files#diff-8b48b2b6c1e1fa25f5d88692710e80fbR110).

Yes, cargo and all processes it spawns afterwards.

sdroege:
What should we do about this one? This is clearly a bug in brew (but they consider it correct behaviour) and we could try to work around it as long as it does not affect other setups.

cmyr:
I am not sure what the next steps are. I can get this working for `cairo-rs-sys` but the same trick wasn't working for cairo proper, which... 😒

sdroege:
Setting `PKG_CONFIG_PATH` to `/usr/local/opt/libffi/lib/pkgconfig` in your environment before calling `cargo` does not work? How does it fail later? Please provide output with `cargo build --verbose` :)

cmyr:
I can do this manually, but I couldn't get it to work via a build script.

sdroege:
Ok for a workaround we'd need to figure out how to set it not only for the currently built crate but for all of them? That seems problematic and probably not possible cleanly.

Independent of all this, did someone report this to homebrew? Even if this is by design, they should then at least also remove all other pkg-config files that depend on libffi.

GuillaumeGomez:
Adding this to the release blockers btw. Very problematic on macOS.

sdroege:
To re-iterate what I said on IRC, I disagree that this should be a release blocker or even something that we must fix. It's clearly a homebrew bug, or rather: it's all working as they intend things to work. That is, not at all without manual intervention. See https://github.com/gtk-rs/cairo/pull/264#issuecomment-501312243

It seems that still nobody bothered to even report this to them, so implementing a working workaround for it is even less likely to happen as it will require quite some effort to get right. If it's possible at all: we somehow need to set `PKG_CONFIG_PATH` not just for one `build.rs` but for all crates somehow and I don't think that's possible (or if it was then I'd even consider that a bug in `cargo` as it breaks the isolation between different crates).

My suggestion here would be to simply document this, or to recommend to not use homebrew as it's known to be broken. And maybe some day someone will also report it to homebrew that their current approach is broken.

nirbheek:
Note that this applies to a number of packages on Homebrew, not just libffi. For instance, it also applies to openssl, pixman, and many others. Search for ["keg-only packages"](https://docs.brew.sh/FAQ#what-does-keg-only-mean).

It's not a Homebrew bug that they don't link it into `/usr/local`, but it's probably a Homebrew bug that they don't add the pkg-config files to the default path.

@cmyr can you try running `brew link libffi` to see if that works? If so, documenting that step should be enough.

sdroege:
See also https://github.com/Homebrew/homebrew-core/issues/40179#issuecomment-495269792 . It's all working as intended

sdroege:
See https://github.com/gtk-rs/sys/pull/141 for another workaround (aka error message).

cmyr:
@nirbheek `brew link libffi` (and `brew link --force libffi`) fail with:

```sh
Warning: Refusing to link macOS-provided software: libffi
```

sdroege:
@GuillaumeGomez This can be closed then

GuillaumeGomez:
@cmyr Thanks anyway!

PR #265, opened 2019-06-13
Run everything through rustfmt

sdroege:
Should be merged after https://github.com/gtk-rs/cairo/pull/260 and needs an update then.


GuillaumeGomez:
This one will fail anyway so I'll just merge it. Thanks!

PR #267, opened 2019-06-13
Run everything through rustfmt

sdroege:
Do **not** merge before https://github.com/gtk-rs/cairo/pull/260


sdroege:
@GuillaumeGomez This can be merged now once CI is green (well, it won't be green because other crates still need fixes)

sdroege:
@GuillaumeGomez Good apart from unrelated failures.

PR #275, opened 2019-07-04
Reexport the Gradient type too.

sdroege:
@GuillaumeGomez backported to the crate branch, which I assume is what you use for releases?


GuillaumeGomez:
No, I'll create a branch from the crate one and cherry-pick the commit. But thanks for the reminder! :)

PR #276, opened 2019-07-13
Replace mem::uninitialized calls

GuillaumeGomez:
Part of gtk-rs/gir#806 (not a regen but same thing).

cc @EPashkin @sdroege

This one should be the last one.

### DIFF
--- a/src/pdf.rs
+++ b/src/pdf.rs
@@ -8,6 +8,7 @@ use std::io;
 use std::mem;
 use std::ops::Deref;
 use std::path::Path;
+use std::ptr;
 
 use enums::PdfVersion;
 #[cfg(any(all(feature = "pdf", feature = "v1_16"), feature = "dox"))]
@@ -50,11 +51,11 @@ impl PdfSurface {
 
     pub fn get_versions() -> impl Iterator<Item = PdfVersion> {
         let vers_slice = unsafe {
-            let mut vers_ptr: *mut ffi::cairo_pdf_version_t = mem::uninitialized();
-            let mut num_vers = 0;
-            ffi::cairo_pdf_get_versions(&mut vers_ptr as _, &mut num_vers as _);
+            let mut vers_ptr = ptr::null_mut();
+            let mut num_vers = mem::MaybeUninit::uninit();
+            ffi::cairo_pdf_get_versions(&mut vers_ptr, num_vers.as_mut_ptr());
 
-            std::slice::from_raw_parts(vers_ptr, num_vers as _)
+            std::slice::from_raw_parts(vers_ptr, num_vers.assume_init() as _)
         };
         vers_slice.iter().map(|v| PdfVersion::from(*v))
     }
--- a/src/ps.rs
+++ b/src/ps.rs
@@ -8,6 +8,7 @@ use std::io;
 use std::mem;
 use std::ops::Deref;
 use std::path::Path;
+use std::ptr;
 
 use enums::PsLevel;
 use ffi;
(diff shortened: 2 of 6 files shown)


EPashkin:
Thanks, 👍 after rust becomes 1.36.0

GuillaumeGomez:
I forgot to update travis once again... Good catch @EPashkin !

src/pdf.rs
```diff
@@ -50,11 +50,11 @@ impl PdfSurface {
 
     pub fn get_versions() -> impl Iterator<Item = PdfVersion> {
         let vers_slice = unsafe {
-            let mut vers_ptr: *mut ffi::cairo_pdf_version_t = mem::uninitialized();
+            let mut vers_ptr = mem::MaybeUninit::<*mut ffi::cairo_pdf_version_t>::uninit();
```
EPashkin: `<*mut ffi::cairo_pdf_version_t>::` seems unneeded here and in other places

GuillaumeGomez: True but explicit doesn't hurt I guess? :)

EPashkin: Agree

sdroege:
On 13 July 2019 19:33:11 EEST, Evgenii Pashkin <notifications@github.com> wrote:
>EPashkin commented on this pull request.
>
>
>
>> @@ -50,11 +50,11 @@ impl PdfSurface {
> 
>     pub fn get_versions() -> impl Iterator<Item = PdfVersion> {
>         let vers_slice = unsafe {
>-            let mut vers_ptr: *mut ffi::cairo_pdf_version_t =
>mem::uninitialized();
>+            let mut vers_ptr = mem::MaybeUninit::<*mut
>ffi::cairo_pdf_version_t>::uninit();

I'd recommend ptr::null_mut() here like we do for all pointers everywhere.

I'll review everything else tomorrow.


-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

EPashkin:
@sdroege, record filled by `cairo_pdf_get_versions`, so IMHO current version is right.

src/pdf.rs
```diff
@@ -50,11 +50,11 @@ impl PdfSurface {
...
             let mut num_vers = 0;
```
sdroege: This one can be `MaybeUninit::uninit()` for consistency with how we do things elsewhere.

src/pdf.rs
sdroege: This one here is a pointer (to a pointer of versions), and we use `ptr::null()` or `ptr::null_mut()` for unitialized pointers elsewhere so it might be better to do that for consistency here too.

src/ps.rs
```diff
@@ -92,11 +92,11 @@ impl PsSurface {
...
     pub fn get_levels() -> impl Iterator<Item = PsLevel> {
         let lvls_slice = unsafe {
-            let mut vers_ptr: *mut ffi::cairo_ps_level_t = mem::uninitialized();
+            let mut vers_ptr = mem::MaybeUninit::<*mut ffi::cairo_ps_level_t>::uninit();
             let mut num_vers = 0;
-            ffi::cairo_ps_get_levels(&mut vers_ptr as _, &mut num_vers as _);
+            ffi::cairo_ps_get_levels(vers_ptr.as_mut_ptr(), &mut num_vers as _);
 
-            std::slice::from_raw_parts(vers_ptr, num_vers as _)
+            std::slice::from_raw_parts(vers_ptr.assume_init(), num_vers as _)
```
sdroege: Same as above

src/svg.rs
```diff
@@ -53,11 +53,11 @@ impl SvgSurface {
...
     pub fn get_versions() -> impl Iterator<Item = SvgVersion> {
         let vers_slice = unsafe {
-            let mut vers_ptr: *mut ffi::cairo_svg_version_t = mem::uninitialized();
+            let mut vers_ptr = mem::MaybeUninit::<*mut ffi::cairo_svg_version_t>::uninit();
             let mut num_vers = 0;
-            ffi::cairo_svg_get_versions(&mut vers_ptr as _, &mut num_vers as _);
+            ffi::cairo_svg_get_versions(vers_ptr.as_mut_ptr(), &mut num_vers as _);
 
-            std::slice::from_raw_parts(vers_ptr, num_vers as _)
+            std::slice::from_raw_parts(vers_ptr.assume_init(), num_vers as _)
```
sdroege: Same as above :)

GuillaumeGomez:
Updated.

sdroege:
:+1:

sdroege:
@GuillaumeGomez all green

PR #280, opened 2019-08-10
Make winapi optional.

RazrFalcon:
Currently, the `winapi` crate is mandatory. But it's very big (100MB). This patch makes it optional by introducing the `win32-surface` feature.

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -13,15 +13,15 @@ extern crate glib_sys as glib_ffi;
 #[cfg(any(feature = "xlib", feature = "dox"))]
 extern crate x11;
 
-#[cfg(windows)]
+#[cfg(feature = "win32-surface")]
 extern crate winapi as winapi_orig;
 
-#[cfg(windows)]
+#[cfg(feature = "win32-surface")]
 pub mod winapi {
     pub use winapi_orig::shared::windef::HDC;
 }
 
-#[cfg(all(not(windows), feature = "dox"))]
+#[cfg(all(feature = "dox", not(feature = "win32-surface")))]
 pub mod winapi {
    use libc::c_void;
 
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Big improvement indeed, thanks a lot!

cc @EPashkin @sdroege

EPashkin:
@RazrFalcon Thanks
There need more cfgs in sys:
https://ci.appveyor.com/project/GuillaumeGomez/cairo-p0df1/builds/26605333/job/pbrdjskstmh8t1j7#L346

RazrFalcon:
@EPashkin It has cfgs already. Not sure what it doesn't like. I've tested it on a local machine first.

Looks like `winapi` should be exported with `dox` on Windows too. Can you clarify that?

RazrFalcon:
I've enabled winapi for `dox` too. Maybe this would help.

cairo-sys-rs/src/lib.rs
```diff
@@ -13,10 +13,10 @@ extern crate glib_sys as glib_ffi;
 #[cfg(any(feature = "xlib", feature = "dox"))]
 extern crate x11;
 
-#[cfg(windows)]
+#[cfg(all(windows, feature = "win32-surface"))]
 extern crate winapi as winapi_orig;
 
-#[cfg(windows)]
+#[cfg(all(windows, feature = "win32-surface"))]
```
EPashkin: IMHO also need to add "dox" here and to previous cfg

EPashkin: No, I wrong, "dox" in Cargo.toml not needed "winapi" as it will fail on nonwindows.
And here need separate module for it
```
#[cfg(all(feature = "dox"))]
pub mod winapi {
    pub enum _HDC {}
        pub type HDC = *mut _HDC;
}
```

EPashkin: Maybe with and `and(.., not(all(windows, feature = "win32-surface"))` but IMHO this unneeded as docs will be generated on unix anyway.

EPashkin:
@RazrFalcon Thanks, 👍

sdroege:
:+1:

PR #284, opened 2019-09-20
Fix invalid cargo key for docs.rs

GuillaumeGomez:
Part of https://github.com/gtk-rs/gir/issues/845

cc @sdroege @EPashkin


EPashkin:
👍

Issue #285, opened 2019-10-04
Add quick explanation of feature flags in documentation


Hofer-Julian:
As mentioned in the title, it would be extremely helpful to introduce the feature flags in the main page of the documentation.

GuillaumeGomez:
This is an excellent suggestion. Interested into doing it? :)

Meanwhile I'm adding this issue for the next release.

Hofer-Julian:
Sure, but I guess I'll have to dig into the code a bit to make sure I write nothing wrong ;)

GuillaumeGomez:
Don't worry, we can help too. I think it'll allow you to discover a few interesting things so it's always nice!

Hofer-Julian:
Great! I'll start to work on it in ~ 2 weeks :)

PR #286, opened 2019-10-06
Include crate features in docs

Hofer-Julian:
Fixes https://github.com/gtk-rs/cairo/issues/285

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,6 +2,63 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+//! # Cairo bindings
+//!
+//! ## Crate features
+//!
+//! ### Default-on features
+//!
+//! * **use_glib** -
+//!   Explanation
+//!
+//! ### Fileformat features
+//!
+//! * **png** -
+//!   Explanation
+//!
+//! * **pdf** -
+//!   Explanation
+//!
+//! * **svg** -
+//!   Explanation
+//!
+//! * **ps** -
+//!   Explanation
+//!
+//! ### Cairo API version features
+//!
+//! * **v1_14** -
+//!   Explanation
+//!
+//! * **v1_16** -
+//!   Explanation
+//!
+//! ### Documentation features
+//!
+//! * **purge-lgpl-docs** -
+//!   Explanation
+//!
+//! * **embed-lgpl-docs** -
+//!   Explanation
+//!
+//! * **dox** -
+//!   Explanation
+//!
+//! ### X window features
+//!
+//! * **xcb** -
+//!   Explanation
+//!
+//! * **xlib** -
+//!   Explanation
+//!
+//! ### Windows API features
+//!
+//! * **win32-surface** -
+//!   Explanation
+//!
+
+
 extern crate cairo_sys as ffi;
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
From my side this pull request would be finished.
If something needs to be changed please let me know.

src/lib.rs
```diff
@@ -4,58 +4,61 @@
...
+//!   This library contains safe Rust bindings for [Cairo](https://www.cairographics.org/).
+//!   It is a part of [Gtk-rs](http://gtk-rs.org/).
+//!
 //! ## Crate features
 //!
 //! ### Default-on features
 //!
 //! * **use_glib** -
-//!   Explanation
+//!   Use [glib](https://gtk-rs.org/docs/glib/)
```
EPashkin: maybe "Use with glib" ?

EPashkin:
@Hofer-Julian Thanks, 👍

src/lib.rs
Hofer-Julian: Good point!

src/lib.rs
```diff
@@ -2,6 +2,66 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+//! # Cairo bindings
+//!
+//!   This library contains safe Rust bindings for [Cairo](https://www.cairographics.org/).
```
GuillaumeGomez: Why the two extra whitespaces at the beginning?

src/lib.rs
```diff
@@ -2,6 +2,66 @@
...
+//!
+//! ### Default-on features
+//!
+//! * **use_glib** -
+//!   Use with [glib](https://gtk-rs.org/docs/glib/)
+//!
+//! ### Fileformat features
+//!
+//! * **png** -
+//!   Reading and writing PNG images
```
GuillaumeGomez: Why putting it on a newline?

src/lib.rs
Hofer-Julian: Readability?
But if the convention differs, I will remove them.

src/lib.rs
Hofer-Julian: No particular reason, I can remove them as well.

src/lib.rs
GuillaumeGomez: It's markdown so they'll get removed. And it makes the doc comment looking strange. It'd be nice to remove then if you don't mind. :)

src/lib.rs
GuillaumeGomez: Yes please!

src/lib.rs
Hofer-Julian: Done!
I removed quite a lot of whitespace.
If you want some back just tell me.

src/lib.rs
GuillaumeGomez: What I meant was to write it this: `* **png** - Reading and writing PNG images`

Hofer-Julian: Done!
It looks indeed better

Hofer-Julian:
I noticed that rust-fmt, made a small change.
Should I remove it?

GuillaumeGomez:
No, don't worry about it. I'm looking at why the doc checker isn't happy.

GuillaumeGomez:
Well whatever, merging. Thanks!

Issue #251, opened 2019-03-14
Surface.create_similar() should return a Result


federicomenaquintero:
As in #141, `cairo_surface_create_similar()` will never return NULL, but it may return a `cairo_surface_t *` with an error status.  I think `Surface::create_similar()` should return a `Result<Surface, Status>`.

sdroege:
Yes, and probably the same for various other functions. Someone should go through them all at some point.

EPashkin:
@federicomenaquintero Thanks for fixing
👍

sdroege:
> @federicomenaquintero Thanks for fixing

This one is not fixed yet :)

federicomenaquintero:
Damn, just got a librsvg bug where we panicked because Surface.create_similar() always returns a Surface :disappointed: 

I've started fixing this.  My first cut is to turn `Surface::from_raw_full() -> Result`, just like we had `ImageSurface::from_raw_full() -> Result`.  The fallout is not bad; it actually makes things like `RecordingSurface::create()` more correct; it returns an Option right now, but cairo_recording_surface_create() will never return NULL, just a surface in an error status.

I have a question about this:

```rust
#[cfg(feature = "use_glib")]
impl FromGlibPtrFull<*mut ffi::cairo_surface_t> for Surface {
    #[inline]
    unsafe fn from_glib_full(ptr: *mut ffi::cairo_surface_t) -> Surface {
        Self::from_raw_full(ptr).unwrap()
    }
}
```

This is.... wrong?  Or did we ever discuss this - the unwrap() seems perilous?

sdroege:
> This is.... wrong? Or did we ever discuss this - the unwrap() seems perilous?

That's correct. If all you want to have returned is a concrete value instead of an Option, then the only thing we can do here is to panic if we get NULL or it doesn't work otherwise.

The Option FromGlibXXX impls would do the right thing and not panic.

EPashkin:
And maybe better add function to `Status` that convert T to Option<T> (near `ensure_valid`) and use it in your other functions

sdroege:
`Result::ok()` you mean?

EPashkin:
@sdroege No, I meant in function that must return Option.

EPashkin:
Oh, ignore this idea, I wrongly understand main issue

federicomenaquintero:
@EPashkin I'm using your `to_result()` idea; thanks for that.  The others are already handled by `Surface.from_raw_full()`, I think.

Issue #141, opened 2017-07-07
cairo_surface_create*() can return errors


federicomenaquintero:
Cairo's surface creation functions can return errors in the form of a valid cairo_surface_t* but whose cairo_surface_status(surf) != CAIRO_STATUS_SUCCESS.

For example, cairo_image_surface_create() can return CAIRO_STATUS_INVALID_SIZE if the passed size is greater than what Pixman supports, or it can return CAIRO_STATUS_NO_MEMORY if allocation fails.

Cairo-rs similarly returns a valid Surface (or ImageSurface) whose status must be checked before proceeding.

Should we return a Result<Surface, Status> (or the corresponding Result<ImageSurface, Status>) instead?

GuillaumeGomez:
Seems like so. Good suggestion!

federicomenaquintero:
Are API-breaking PRs accepted? :)

GuillaumeGomez:
Yup, we already have a few. The next release will be a "middle" one.

federicomenaquintero:
I have a question about possible APIs.

Let's say you do
```
let raw_surface = ffi::cairo_image_surface_create(Format::ARgb32, 50000, 50000);
```
Cairo will return a surface with a status of CAIRO_STATUS_INVALID_SIZE, as pixman only supports sizes up to 32767.  For a valid size, it could return CAIRO_STATUS_NO_MEMORY.  Also, `cairo_image_surface_create_for_data()` could return CAIRO_STATUS_INVALID_STRIDE, etc.

At first I was thinking that the cairo-rs functions that create surfaces could simply return `Result<ImageSurface, Status>`.  However, the functions that convert raw pointers into Rust structures should simply return an `ImageSurface` which possibly have a `status != Success`.  This includes `ImageSurface::from_raw_full()` and the corresponding `impl FromGlibPtrFull<*mut ffi::cairo_surface_t> for ImageSurface`.  The rationale is that we can keep a Rust-like API for the creation functions, but for data that comes from elsewhere, we just pass it through the wrapper and let the user check for statuses.  That is, it would look like

```
impl ImageSurface {
    pub fn create(format: Format, width: i32, height: i32) -> Result<ImageSurface, Status> { ... }
    pub fn create_for_data<F>(data: Box<[u8]>, free: F, format: Format, width: i32, height: i32, 
                              stride: i32) -> Result<ImageSurface, Status> { ... }

[...]

PR #287, opened 2019-10-09
(#251): Surface::create_similar() and friends should return a Result

federicomenaquintero:
Cairo's surface creation functions never return NULL; instead they
always return a surface, but it may be in an error state.  In #141 we
started making the binding functions return Result for this; some
returned the plain FooSurface type, some others Option<FooSurface>.

This makes the following functions return Result<FooSurface, Status>

  Surface::create_similar()
  Surface::create_similar_image()
  RecordingSurface::create()
  Device.surface_create()
  Device.surface_create_for_target()

The foundation for all of this is that Surface::from_raw_full() now
also returns Result<FooSurface, Status>.  This is to make things
consistent with ImageSurface::from_raw_full().  Analogously, we now
have RecordingSurface::from_raw_full() that also returns Result.

Fixes https://github.com/gtk-rs/cairo/issues/251


federicomenaquintero:
Bleh, forgot to check  the build with features.  I'll fix this tomorrow.

federicomenaquintero:
Fixed the features.

Also, went along and fixed all the PdfSurface functions that may set &self in an error state.

EPashkin:
AppVeyor's errors are legit

src/pdf.rs
```diff
@@ -170,7 +176,7 @@ impl FromGlibPtrFull<*mut ffi::cairo_surface_t> for PdfSurface {
     #[inline]
     unsafe fn from_glib_full(ptr: *mut ffi::cairo_surface_t) -> PdfSurface {
         Self {
-            inner: Surface::from_raw_full(ptr),
+            inner: Surface::from_raw_full(ptr).unwrap(),
```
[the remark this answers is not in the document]
federicomenaquintero: Uuh, I think this needs a treatment similar to ImageSurface, with TryFrom.

src/pdf.rs
```diff
@@ -60,20 +59,22 @@ impl PdfSurface {
         vers_slice.iter().map(|v| PdfVersion::from(*v))
     }
 
-    pub fn restrict(&self, version: PdfVersion) {
+    pub fn restrict(&self, version: PdfVersion) -> Result<(), Status> {
         unsafe {
             ffi::cairo_pdf_surface_restrict_to_version(self.inner.to_raw_none(), version.into());
         }
+        self.status().to_result(())
```
sdroege: What does this mean exactly? Is the surface still usable afterwards or is it completely broken? :)

sdroege:
Looks good to me generally, thanks for the work :)

src/pdf.rs
federicomenaquintero: Most cairo_pdf_surface_*() functions start by calling [`_extract_pdf_surface()`](https://gitlab.freedesktop.org/cairo/cairo/blob/52a7c79f/src/cairo-pdf-surface.c#L620-665), which can set the surface in an error state.

Afterwards... not all of cairo_surface_*() check for the surface->status.  It's not completely unusable if in error, but some functions *will* still work ¯\_(ツ)_/¯

federicomenaquintero:
Let me see if I can fix the TryFrom thing - to make it consistent with other concrete surfaces that wrap the abstract one.

sdroege:
Any updates here?

sdroege:
@federicomenaquintero we'd like to get a release out next weekend. Any chance this can be ready by then? :)

federicomenaquintero:
I'll try to work on this this week.  Need to remind myself what's left to be done :)

sdroege:
@federicomenaquintero Thanks, let us know when it's time for another review or if you're stuck somewhere :)

federicomenaquintero:
I think it's ready.

The next step is probably to take all the functions that now all look alike - FooSurface::try_from(Surface); the to_glib ones - and make a macro for them or something.

federicomenaquintero:
Argh, QuartzSurface. I'll deal with it.

federicomenaquintero:
Now all derived surfaces are done with a macro.  This even buys us some methods that were missing from the old implementations.

sdroege:
Looks good to me, thanks! This cleans up everything quite nicely.

The CI failure in the examples is expected, so this is IMHO good to go once all jobs have failed only when building the examples and not before that @GuillaumeGomez

sdroege:
@GuillaumeGomez Please trigger CI once https://github.com/gtk-rs/examples/pull/271 is merged and then this can go in IMHO

sdroege:
@GuillaumeGomez @federicomenaquintero minimum rust version in the CI needs to be updated to 1.39 or it will fail :) Can one of you push a fix for that to this branch?

sdroege:
Now only misses a `cargo fmt` run. @GuillaumeGomez? :)

federicomenaquintero:
Uuuuh, what changed that requires Rust 1.39?  I mean, I need to update things in Suse, anyway, but just to know.

federicomenaquintero:
Thanks for the merge!

GuillaumeGomez:
async/await in other crates. That wouldn't make sense to have cairo handle its version on its own.

federicomenaquintero:
Got it.  Sounds good, thanks!

PR #288, opened 2019-10-20
Improve cairo library documentation

GuillaumeGomez:


GuillaumeGomez:
I don't understand the failure in here but whatever...

EPashkin:
@GuillaumeGomez Failure on doc_check, because src/lib.rs not ignored in purge-lgpl-docs

GuillaumeGomez:
Nice catch! I was looking in the wrong direction. :)

PR #294, opened 2019-11-28
SvgSurface: make filename param optional

### LINKED ISSUE #293: SvgSurface path should be an Option
Per the docs in https://www.cairographics.org/manual/cairo-SVG-Surfaces.html the filename param of  `cairo::SvgSurface::new` should be optional see

> filename | a filename for the SVG output (must be writable), NULL may be used to specify no output. This will generate a SVG surface that may be queried and used as a source, without generating a temporary file.

bilelmoussaoui:
Per the docs https://www.cairographics.org/manual/cairo-SVG-Surfaces.html
This allows drawing parts of a SVG file without saving it to a tmp file.

I'm not sure if what I did is right or not :P 

Fixes #293

### DIFF
--- a/src/svg.rs
+++ b/src/svg.rs
@@ -2,19 +2,21 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+#[cfg(any(all(feature = "svg", feature = "v1_16"), feature = "dox"))]
+use enums::SvgUnit;
+use enums::{Status, SurfaceType, SvgVersion};
+use ffi;
 use std::convert::TryFrom;
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::io;
 use std::mem;
 use std::ops::Deref;
+#[cfg(not(windows))]
+use std::os::unix::prelude::*;
 use std::path::Path;
 use std::ptr;
 
-#[cfg(any(all(feature = "svg", feature = "v1_16"), feature = "dox"))]
-use enums::SvgUnit;
-use enums::{Status, SurfaceType, SvgVersion};
-use ffi;
 use surface::Surface;
 
 #[cfg(feature = "use_glib")]
@@ -33,13 +35,37 @@ impl SvgVersion {
 declare_surface!(SvgSurface, SurfaceType::Svg);
 
 impl SvgSurface {
-    pub fn new<P: AsRef<Path>>(width: f64, height: f64, path: P) -> Result<SvgSurface, Status> {
-        let path = path.as_ref().to_string_lossy().into_owned();
-        let path = CString::new(path).unwrap();
+    pub fn new<P: AsRef<Path>>(
+        width: f64,
+        height: f64,
+        path: Option<P>,
+    ) -> Result<SvgSurface, Status> {
+        #[cfg(not(windows))]
+        let path = path.map(|p| {
+            CString::new(p.as_ref().as_os_str().as_bytes()).expect("Invalid path with NULL bytes")
+        });
+        #[cfg(windows)]
+        let path = path.map(|p| {
+            let path_str = p
+                .as_ref()
+                .to_str()
+                .expect("Path can't be represented as UTF-8")
+                .to_owned();
+            if path_str.starts_with("\\\\?\\") {
+                CString::new(path_str[4..].as_bytes())
+            } else {
+                CString::new(path_str.as_bytes())
+            }
+            .expect("Invalid path with NUL bytes")
+        });
 
         unsafe {
             Ok(Self(Surface::from_raw_full(
-                ffi::cairo_svg_surface_create(path.as_ptr(), width, height),
+                ffi::cairo_svg_surface_create(
+                    path.as_ref().map(|p| p.as_ptr()).unwrap_or(ptr::null()),
+                    width,
+                    height,
+                ),
             )?))
         }
     }
@@ -85,7 +111,7 @@ impl SvgSurface {
 mod test {
     use super::*;
     use context::*;
-    use tempfile::tempfile;
+    use tempfile::{tempfile, NamedTempFile};
 
     fn draw(surface: &Surface) {
         let cr = Context::new(surface);
@@ -130,9 +156,16 @@ mod test {
     }
 
     #[test]
-    #[cfg(unix)]
+    fn without_file() {
+        let surface = SvgSurface::new(100., 100., None::<&Path>).unwrap();
+        draw(&surface);
+        surface.finish();
+    }
+
+    #[test]
     fn file() {
-        let surface = SvgSurface::new(100., 100., "/dev/null").unwrap();
+        let file = NamedTempFile::new().expect("tempfile failed");
+        let surface = SvgSurface::new(100., 100., Some(&file.path())).unwrap();
         draw(&surface);
         surface.finish();
     }


bilelmoussaoui:
@GuillaumeGomez Any idea why the CI fails with 
```
error[E0599]: no method named `to_glib_none` found for type `std::option::Option<&std::string::String>` in the current scope

  --> src/svg.rs:44:61

   |

44 |                 ffi::cairo_svg_surface_create(path.as_ref().to_glib_none().0, width, height),

   |                                                             ^^^^^^^^^^^^ method not found in `std::option::Option<&std::string::String>`

```
It builds and works fine locally here.

GuillaumeGomez:
Try to change your code as follow:

```rust
path.map(|s| s.as_str()).to_glib_none().0
```

bilelmoussaoui:
My bad! I was running the tests without `--features svg`. It should pass this time. Thanks!

bilelmoussaoui:
It fails because tests are run without `--features use_glib` :man_facepalming:

GuillaumeGomez:
Oh then it depends: is it supposed to be run even without glib or not? If so, please don't use `to_glib` calls.

bilelmoussaoui:
It doesn't need to. In such case, what I'm supposed to do?

GuillaumeGomez:
Like this:

```rust
let c_str = CString::new(c_str).unwrap();
// then:
c_str.as_ptr();
```

src/svg.rs
```diff
@@ -33,13 +33,15 @@ impl SvgVersion {
...
-    pub fn new<P: AsRef<Path>>(width: f64, height: f64, path: P) -> Result<SvgSurface, Status> {
-        let path = path.as_ref().to_string_lossy().into_owned();
-        let path = CString::new(path).unwrap();
-
+    pub fn new<P: AsRef<Path>>(
+        width: f64,
+        height: f64,
+        path: Option<P>,
+    ) -> Result<SvgSurface, Status> {
+        let path = path.map(|p| p.as_ref().to_string_lossy().into_owned());
```
sdroege:
```suggestion
        let path = path.map(|p| CString::new(path.as_ref().as_os_str().as_bytes()));
        let path = path.map(|p| p.as_ptr()).unwrap_or(ptr::null());
```

Ideally you'd do more than that though (Windows-specific special case...): https://github.com/gtk-rs/glib/blob/faff6573602daec764584030e959f7d28a0aef81/src/translate.rs#L421-L457

src/svg.rs
```diff
@@ -2,19 +2,21 @@
...
+use enums::{Status, SurfaceType, SvgVersion};
+use ffi;
 use std::convert::TryFrom;
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::io;
 use std::mem;
 use std::ops::Deref;
+#[cfg(not(windows))]
+use std::os::unix::prelude::*;
```
sdroege: Without a windows case this is not going to compile on Windows. Looks ok for UNIX though.

sdroege:
lgtm

src/svg.rs
```diff
@@ -33,13 +35,37 @@ impl SvgVersion {
...
+            let path = path.map(|p| {
+                CString::new(p.as_ref().as_os_str().as_bytes())
+                    .expect("Invalid path with NULL bytes")
+            });
+            path.map(|p| p.as_ptr()).unwrap_or(ptr::null())
+        };
+        #[cfg(windows)]
+        let path = {
+            // Based on GLib path to C
+            let path_str = path
```
sdroege: This assumes that `path` is not an `Option` and already a `Path`. Needs the `map()`, `as_ref()` part from above too

src/svg.rs
```diff
@@ -33,13 +35,39 @@ impl SvgVersion {
...
+        #[cfg(windows)]
+        let path = {
+            // Based on GLib path to C
+            let path_str = path.map(|p| {
+                p.as_ref()
+                    .to_str()
+                    .expect("Path can't be represented as UTF-8")
+                    .to_owned()
+            });
+            if path_str.starts_with("\\\\?\\") {
```
sdroege: This and the following lines need to stay inside the `map` closure. You still have an `Option` here. Later you need to also get the pointer of the whole thing

src/svg.rs
```diff
@@ -33,13 +35,39 @@ impl SvgVersion {
...
+        height: f64,
+        path: Option<P>,
+    ) -> Result<SvgSurface, Status> {
+        #[cfg(not(windows))]
+        let path = {
+            let path = path.map(|p| {
+                CString::new(p.as_ref().as_os_str().as_bytes())
+                    .expect("Invalid path with NULL bytes")
+            });
+            path.map(|p| p.as_ptr()).unwrap_or(ptr::null())
```
sdroege: The path will be freed once this block is done, and then you store a dangling pointer in path in the outer scope. Just return the `Option<CString>` from this block, and do the `as_ptr()` part for both Windows/UNIX just before calling the C function.

sdroege:
Looks good for real now :)

GuillaumeGomez:
Then I'll merge for real once CI passed. :laughing:

EPashkin:
@bilelmoussaoui Thanks, 👍

EPashkin:
@sdroege, @GuillaumeGomez Maybe we need to make these function public someday
https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L422-L486

GuillaumeGomez:
@EPashkin Oh indeed! Do you mind opening an issue for it please? So it can be done for the next release. :)

sdroege:
That would not help here as otherwise cairo would have to always depend on glib

EPashkin:
@sdroege Agree about this case, but IMHO here also better add separate `path_to_c` and use with `map` to make code more clear

bilelmoussaoui:
> @sdroege Agree about this case, but IMHO here also better add separate `path_to_c` and use with `map` to make code more clear

I don't see a valid reason to do that as the function won't be used anywhere else but I agree it makes that part of code less readable. 

The failing tests are not related to this change I think.

EPashkin:
@bilelmoussaoui Why you think that it not related? 
```
failures:

---- svg::test::file stdout ----

thread 'svg::test::file' panicked at 'called `Result::unwrap()` on an `Err` value: "error while writing to output stream"', src/libcore/result.rs:1189:5

failures:

    svg::test::file

test result: FAILED. 39 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

EPashkin:
IMHO test with `None` also will be good

bilelmoussaoui:
Because I had that issue once, tried to reproduce it and it was gone the next time I ran the tests. Yeah a test with None will be good. Will add that.

src/svg.rs
```diff
@@ -132,7 +158,7 @@ mod test {
     #[test]
     #[cfg(unix)]
     fn file() {
-        let surface = SvgSurface::new(100., 100., "/dev/null").unwrap();
+        let surface = SvgSurface::new(100., 100., Some("/dev/null")).unwrap();
```
EPashkin: Seems this test need be just `None`, and `Some` need use `tempfile`

bilelmoussaoui: Yeah, this is probably the reason why it fails.

src/svg.rs
```diff
@@ -130,9 +156,16 @@ mod test {
...
+    fn without_file() {
+        let surface = SvgSurface::new(100., 100., None::<&Path>).unwrap();
+        draw(&surface);
+        surface.finish();
+    }
+
+    #[test]
     fn file() {
-        let surface = SvgSurface::new(100., 100., "/dev/null").unwrap();
+        let file = NamedTempFile::new().expect("tempfile failed");
```
EPashkin: Why not `tempfile()` ?

sdroege:
Still fails to write the output file, on Windows and macOS but not on Linux.

EPashkin:
On Windows `SvgSurface::new` not work even for direct path

EPashkin:
Old version works on `let surface = SvgSurface::new(100., 100., "_.tmp").unwrap();`,
I try to find problem

src/svg.rs
```diff
@@ -33,13 +35,37 @@ impl SvgVersion {
...
+                CString::new(path_str.as_bytes())
+            }
+            .expect("Invalid path with NUL bytes")
+        });
 
         unsafe {
             Ok(Self(Surface::from_raw_full(
-                ffi::cairo_svg_surface_create(path.as_ptr(), width, height),
+                ffi::cairo_svg_surface_create(
+                    path.map(|p| p.as_ptr()).unwrap_or(ptr::null()),
```
EPashkin: Windows works with `path.as_ref().map(|p| p.as_ptr()).unwrap_or(ptr::null()),`

sdroege: Ok yeah, not doing `as_ref()` will consume the `CString` here and we have a dangling pointer again. Not just for Windows.

src/svg.rs
bilelmoussaoui: Because I couldn't find a way to get a `Path` from `tempfile()`

EPashkin: Your right

EPashkin:
CI passed

sdroege:
Great, let's get it in then :)

GuillaumeGomez:
Thanks!

PR #300, opened 2020-01-30
Prevent linking libraries with "dox" feature to avoid documentation f…

EPashkin:
…ailure

Found in https://github.com/gtk-rs/glib/issues/582

It seems single sys crate in gtk-rs, that don't have these changes.

cc @GuillaumeGomez, @sdroege


sdroege:
Makes sense!

GuillaumeGomez:
Arf, we missed one! Thanks a lot @EPashkin !

PR #303, opened 2020-02-07
Remove deprecated Error::description

EPashkin:
Fix
```
warning: use of deprecated item 'std::error::Error::description': use the Display impl or to_string()
  --> D:/eap/rust/0/cairo\src\error.rs:69:37
   |
69 |             IoError::Io(ref e) => e.description(),
   |                                     ^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

cc @GuillaumeGomez , @sdroege

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,7 @@ git = "https://github.com/gtk-rs/sys"
 [dependencies]
 libc = "0.2"
 bitflags = "1.0"
+thiserror = "1.0.10"
 
 [dev-dependencies]
 tempfile = "3.0"
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,13 +3,14 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use enums::Status;
-use std::error::Error;
-use std::fmt;
 use std::io;
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum BorrowError {
+    #[error("BorrowError::Cairo({0})")]
     Cairo(Status),
+    #[error("BorrowError::NonExclusive")]
     NonExclusive,
 }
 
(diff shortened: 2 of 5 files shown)


GuillaumeGomez:
We'll have to do it in gir as well.

EPashkin:
There is no `description` in gir

GuillaumeGomez:
We don't generate code using `description`? I was sure we did...

EPashkin:
CI ended with only failed osx rust beta

sdroege:
We also implement the `description()` method in lots of other `std::error::Error` impls in the other crates. Removing it has the disadvantage that old code (still using it) would print nothing useful anymore, so I'm not sure removing it is a good idea at this point. It's deprecated since 1.41 only!

In gstreamer-rs I decided to side-step this whole topic and let the `std::error::Error` impls be derived automatically be the `thiserror` crate. Removed lots of boring boilerplate and works around this problem (currently it still generates it but silences the warning, at some later point it will stop generating it when the whole ecosystem moved on and we don't need to change anything).

dtolnay:
> In gstreamer-rs I decided to side-step this whole topic and let the `std::error::Error` impls be derived automatically by the `thiserror` crate. Removed lots of boring boilerplate and works around this problem (currently it still generates it but silences the warning, at some later point it will stop generating it when the whole ecosystem moved on and we don't need to change anything).

FYI thiserror has never generated description methods.

sdroege:
> FYI thiserror has never generated description methods.

Oh :sweat_smile: I should've checked the MR switching to `thiserror` in more detail, I guess.

But that's also good information. Nobody complained about it in the last months, so simply dropping `description()` is probably not a problem for the next release.

@EPashkin please reopen and then let's get it in :) Unless you want to move to `thiserror` for getting rid of some boilerplate while at it.

EPashkin:
Se

sdroege:
The `Display` impls could be a bit more descriptive. They're right now basically `Debug` impls, not very user-friendly. Another reason to enforce reasonable behaviour by having this stuff be autogenerated ;)

EPashkin:
Yes, I will try to use `thiserror`, it only generates `Error, `Display` and `From`

Oftopic: @sdroege what you think about https://crates.io/crates/snafu ?

sdroege:
> Oftopic: @sdroege what you think about https://crates.io/crates/snafu ?

It's similar but does a bit more from what I understand. `thiserror` would only do what you would write manually anyway.

EPashkin:
Updated

EPashkin:
Updated.
@sdroege, @GuillaumeGomez better messages are welcome.

src/error.rs
```diff
@@ -19,61 +20,16 @@ impl From<Status> for BorrowError {
...
-    }
-}
-
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum IoError {
+    #[error("IO operation result: {0}")]
     Cairo(Status),
-    Io(io::Error),
+    #[error("{0}")]
```
sdroege:
```suggestion
    #[error("IO error: {0}")]
```

maybe?

src/error.rs
```diff
@@ -19,61 +20,16 @@ impl From<Status> for BorrowError {
...
-
-    fn cause(&self) -> Option<&dyn Error> {
-        None
-    }
-}
-
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum IoError {
+    #[error("IO operation result: {0}")]
```
sdroege:
```suggestion
    #[error("Cairo error: {0}")]
```

src/error.rs
```diff
@@ -3,13 +3,14 @@
...
 use enums::Status;
-use std::error::Error;
-use std::fmt;
 use std::io;
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum BorrowError {
+    #[error("Borrow error status:{0}")]
```
sdroege:
```suggestion
    #[error("Failed to borrow with Cairo error: {0}")]
```

maybe?

src/error.rs
EPashkin: Updated

src/error.rs
```diff
@@ -3,13 +3,14 @@
...
     Cairo(Status),
+    #[error("Can't take it exclusively")]
```
sdroege:
```suggestion
    #[error("Can't get exclusive access")]
```

EPashkin:
@sdroege Thanks, I applied all suggestions

Cargo.toml
```diff
@@ -60,6 +60,7 @@ git = "https://github.com/gtk-rs/sys"
 [dependencies]
 libc = "0.2"
 bitflags = "1.0"
+thiserror = "1.0.10"
```
GuillaumeGomez: So we're adding new proc-macro dependencies? :'(

EPashkin: @GuillaumeGomez Sorry, I totally forgot about this comment.
You don't want this dependence?
I can only remove `fn description` and change error texts in `Display`

GuillaumeGomez: It's always advantages versus downsides. For me, it doesn't bring enough to justify it but @sdroege might disagree on this.

sdroege: It removes some boring boilerplate and does not really add a lot dependency-wise. We depend on all the dependencies apart from `thiserror` already IIRC (in glib/gio).

In case of switching over `gstreamer-rs` to `thiserror`, it also solved an annoying bug that nobody noticed in the manual implementation (infinite recursion in the `Display` impl, indirectly calling into itself).

GuillaumeGomez: Then it's all good.

GuillaumeGomez:
Thanks @EPashkin !

PR #304, opened 2020-02-14
Bugfix: RectangleList's Debug implementation has an infinite recursion

manuthambi:
Use .deref() to ensure we print the dereferenced value.


GuillaumeGomez:
Nice catch, thanks!

sdroege:
Indeed, thanks! @GuillaumeGomez CI is green, and we should do a bugfix release for this ideally :)

GuillaumeGomez:
Sure. Let's go a release!

manuthambi:
Thanks for the quick release!

PR #313, opened 2020-04-04
Add missing Surface::get_device method

GuillaumeGomez:
Fixes #312 

cc @sdroege @EPashkin

PR #314, opened 2020-04-04
Replace bad code

GuillaumeGomez:
As suggested by @psychon (thanks!) in #313.

cc @EPashkin @sdroege

### DIFF
--- a/src/surface.rs
+++ b/src/surface.rs
@@ -157,9 +157,7 @@ impl Surface {
             if device.is_null() {
                 None
             } else {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
@psychon Well, bad code is bad code. Just realized that maybe you wanted to write it? If you prefer I can close this PR so you can open one (you're the one who spotted the issue after all...).

psychon:
I pass. Thanks for doing the work I was and am too lazy to do. :-)

GuillaumeGomez:
Ok :laughing:

Thanks a lot for telling us we did bad code!

Issue #320, opened 2020-05-28
Implement Error trait for cairo::Success


bilelmoussaoui:
I will send a merge request for this one

sdroege:
And change it to `cairo::Error` with only the error cases, and convert to `Result<T, cairo::Error>` where current `cairo::Status::Success` maps to `Ok` and the other things to `Err` :)

bilelmoussaoui:
I have a few questions:
- I noticed a lot of `ensure_status()` which just get the status of  X and calls `ensure_valid()` which itself just checks if the `Status` is `Success` or not. I don't think it makes sense to keep any of those around any more, does it? 
- the `ensure_status()` is called at unexpected places sometimes, where we assume we will not have a Result<T, cairo::Error> at all, for example https://github.com/gtk-rs/cairo/blob/master/src/font/font_options.rs#L41. Does it make more sense to just convert to a Result<T, cairo::Error> there as well?
- a bunch of cairo objects exposes a `pub fn status(&self) -> Status`, does it make sense to still expose that? it might be useful in some weird use cases but doesn't look right in Rust. We should probably remove that and just return a Result whenever something might fail.

My questions are not super clear, sorry for that :)

sdroege:
>     * I noticed a lot of `ensure_status()` which just get the status of  X and calls `ensure_valid()` which itself just checks if the `Status` is `Success` or not. I don't think it makes sense to keep any of those around any more, does it?

I guess those could still stay around as `Result<(), cairo::Error>`... not sure why they exist at all, shouldn't operations that cause things to fail return a `Result` instead? But that needs some more investigation so I would do the removal of those as a second step.

>     * the `ensure_status()` is called at unexpected places sometimes, where we assume we will not have a Result<T, cairo::Error> at all, for example https://github.com/gtk-rs/cairo/blob/master/src/font/font_options.rs#L41. Does it make more sense to just convert to a Result<T, cairo::Error> there as well?

I assume creating a `FontOptions` can fail? As is this is a bit useless, it should either not check it (-> it can't fail) or it should return a `Result` like you suggest.

>     * a bunch of cairo objects exposes a `pub fn status(&self) -> Status`, does it make sense to still expose that? it might be useful in some weird use cases but doesn't look right in Rust. We should probably remove that and just return a Result whenever something might fail.

I agree, but same as with the (public!) `ensure_status()` I think this needs some more research though, same as above.

Does this help?

bilelmoussaoui:
[...]

PR #321, opened 2020-06-03
remove unneded status checking

bilelmoussaoui:
Per the docs that I attached on each commit, these functions don't modify the status and so shouldn't fail. This is an effort to implement #320

### DIFF
--- a/src/font/font_face.rs
+++ b/src/font/font_face.rs
@@ -26,16 +26,14 @@ pub struct FontFace(ptr::NonNull<ffi::cairo_font_face_t>);
 
 impl FontFace {
     pub fn toy_create(family: &str, slant: FontSlant, weight: FontWeight) -> FontFace {
-        let font_face: FontFace = unsafe {
+        unsafe {
             let family = CString::new(family).unwrap();
             FontFace::from_raw_full(ffi::cairo_toy_font_face_create(
                 family.as_ptr(),
                 slant.into(),
                 weight.into(),
             ))
-        };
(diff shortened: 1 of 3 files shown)


src/font/font_face.rs
```diff
@@ -26,16 +26,14 @@ pub struct FontFace(ptr::NonNull<ffi::cairo_font_face_t>);
 
 impl FontFace {
     pub fn toy_create(family: &str, slant: FontSlant, weight: FontWeight) -> FontFace {
```
sdroege, line 28: What will happen here if you pass an invalid font family?

bilelmoussaoui: Hmm I somehow completely forgot to ask myself that ^^

bilelmoussaoui:
So I have checked the source code of cairo this time, and all of these functions might fail. In such case they should return a Result. I will do that change already :)

sdroege:
> and all of these functions

*all*, including for example `set_antialias()`?!

bilelmoussaoui:
> `set_antialias`

It seems https://github.com/freedesktop/cairo/blob/master/src/cairo.c#L1104-L1111

sdroege:
Right, so if we have everything return a `Result` then the cairo API will be really hard to use...

Can it fail in any reasonable circumstances? If not then this should stay a panic

bilelmoussaoui:
aha Indeed, panic is what I completely missed. I'm closing this one and I have just made the `Result<T, Status>` to `Result<T, Error>` switch

PR #322, opened 2020-06-03
errors: create an Error type and use that instead of Status

### LINKED ISSUE #320: Implement Error trait for cairo::Success

bilelmoussaoui:
fixes #320


src/enums.rs
```diff
@@ -275,11 +276,52 @@ impl Status {
         }
     }
 
-    pub(crate) fn to_result<T>(self, obj: T) -> Result<T, Self> {
-        if self == Status::Success {
-            Ok(obj)
-        } else {
-            Err(self)
+    pub(crate) fn to_result<T>(self, obj: T) -> Result<T, error::Error> {
```
sdroege, line 279: I think the `Status` type should disappear from the public API

src/error.rs
```diff
@@ -1,35 +1,136 @@
...
+    FreetypeError,
+    Win32GdiError,
+    LastStatus,
+    Unknown,
 }
 
-impl From<Status> for BorrowError {
-    fn from(status: Status) -> Self {
-        BorrowError::Cairo(status)
+impl fmt::Display for Error {
```
sdroege, line 56: You can also just derive `Display` via `thiserror` (and we should give them better strings than what we have below...)

src/error.rs
```diff
@@ -1,35 +1,136 @@
...
+            }
+        )
     }
 }
 
 #[derive(Error, Debug)]
 pub enum IoError {
     #[error("Cairo error: {0}")]
-    Cairo(Status),
+    Cairo(Error),
```
sdroege, line 113:
```suggestion
    Cairo(#[from] Error),
```

src/error.rs
```diff
@@ -1,35 +1,136 @@
...
-        IoError::Cairo(status)
+impl From<Error> for IoError {
+    fn from(error: Error) -> Self {
+        IoError::Cairo(error)
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum BorrowError {
+    #[error("Failed to borrow with Cairo error:{0}")]
```
sdroege, line 126:
```suggestion
    #[error("Failed to borrow with Cairo error: {0}")]
```

src/error.rs
bilelmoussaoui: I'm getting 
```
error[E0658]: attributes on expressions are experimental
   --> src/error.rs:134:28
    |
134 |         BorrowError::Cairo(#[from] error)
    |                            ^^^^^^^
    |
    = note: see issue #15701 <https://github.com/rust-lang/rust/issues/15701> for more information
    = help: add `#![feature(stmt_expr_attributes)]` to the crate attributes to enable

```

sdroege: How does it work one line below then? Maybe `#[from] error::Error`, maybe it needs a path...

bilelmoussaoui:
> Looks good to me in general, but IIRC you had an `unwrap()` somewhere that should actually be returned as an error?

That was for surface.create_similar, it returned Status which doesn't implement std::error::Error
 
> Also you might want to consider make the `ensure_status()` function to not panic but to return a `Result`, and do the panicking via `unwrap()` in the caller (or even better `expect()` to give more context). That way it's clearer where things went wrong.

Sounds a cleaner solution. Especially that I'm interested in removing the Status enum altogether.  I have an implementation detail question.
If I have this function for example ```ffi::cairo_pattern_get_surface(  self.pointer,  &mut surface_ptr )``` that returns a `cairo_surface_t` how can I turn that directly into something that's a Result? I would have to match the status manually unless there's Status type or there's some magic I haven't heard of yet?

Thanks as always :)

src/error.rs
bilelmoussaoui: It only works if I remove the `From<Error> for BorrowError` & `From<Error> for IoError` implementations

sdroege: Well that makes sense, the attribute is providing exactly those `From` impls (and marking the contained error as source) :)

bilelmoussaoui:
> And now you can change it to return `Error` and all is good?

Yes, everything works fine from that side of things

> > If I have this function for example `ffi::cairo_pattern_get_surface( self.pointer, &mut surface_ptr )` that returns a `cairo_surface_t` how can I turn that directly into something that's a Result? I would have to match the status manually unless there's Status type or there's some magic I haven't heard of yet?
> 
> ```rust
> let status = ffi::cairo_pattern_get_surface(self.pointer, &mut surface_ptr);
> some_function(status, || convert_surface_ptr_to_return_value)
> ```
> 
> Or what do you mean?

I meant, how I can replace the `to_result` function that I had in `Status` now that I have a `cairo_status_t` ?

src/error.rs
bilelmoussaoui: Aha, makes sense now :)

sdroege:
> I meant, how I can replace the `to_result` function that I had in `Status` now that I have a `cairo_status_t` ?

Have some standalone function that does the same thing :) No need to have that on some type

src/device.rs
```diff
@@ -120,8 +117,11 @@ impl Device {
     }
 
     // Maybe improve this API?
-    pub fn acquire(&self) -> Status {
-        unsafe { Status::from(ffi::cairo_device_acquire(self.to_raw_none())) }
+    pub fn acquire(&self) -> Result<(), Error> {
```
sdroege, line 120: Indeed, let's return a value here that calls `release()` on `Drop`. The way it is now is asking for problems

src/error.rs
```diff
@@ -1,35 +1,214 @@
...
 use thiserror::Error;
 
 #[derive(Error, Debug)]
-pub enum BorrowError {
-    #[error("Failed to borrow with Cairo error:{0}")]
-    Cairo(Status),
-    #[error("Can't get exclusive access")]
-    NonExclusive,
+pub enum Error {
+    #[error("Cairo: No Memory")]
```
sdroege, line 11: Why the `"Cairo: "` prefix everywhere?

src/error.rs
```diff
@@ -1,35 +1,214 @@
...
 use std::io;
 use thiserror::Error;
 
 #[derive(Error, Debug)]
-pub enum BorrowError {
-    #[error("Failed to borrow with Cairo error:{0}")]
-    Cairo(Status),
-    #[error("Can't get exclusive access")]
-    NonExclusive,
+pub enum Error {
```
sdroege, line 10: `#[non_exhaustive]`

src/error.rs
```diff
@@ -1,35 +1,214 @@
...
+    JBig2GlobalMissing,
+    #[error("Cairo: PNG Error")]
+    PngError,
+    #[error("Cairo: Freetype Error")]
+    FreetypeError,
+    #[error("Cairo: Win32Gdi Error")]
+    Win32GdiError,
+    #[error("Cairo: LastStatus")]
+    LastStatus,
+    #[error("Cairo: Uknown {0}")]
```
sdroege, line 95:
```suggestion
    #[error("Cairo: Unknown {0}")]
```

src/error.rs
```diff
@@ -1,35 +1,214 @@
...
+    __Unknown(i32),
```
sdroege, line 96: Maybe `#[doc(hidden)]` this one

src/patterns.rs
```diff
@@ -360,32 +356,32 @@ impl Mesh {
...
     pub fn curve_to(&self, x1: f64, y1: f64, x2: f64, y2: f64, x3: f64, y3: f64) {
         unsafe { ffi::cairo_mesh_pattern_curve_to(self.pointer, x1, y1, x2, y2, x3, y3) }
-        self.ensure_status();
+        self.status().expect("Cairo: Failed to curve to");
     }
 
     pub fn set_control_point(&self, corner: MeshCorner, x: f64, y: f64) {
         unsafe { ffi::cairo_mesh_pattern_set_control_point(self.pointer, corner.into(), x, y) }
-        self.ensure_status();
+        self.status().expect("Cairo: Failed to set control point");
```
sdroege, line 384: Also why here all the `"Cairo:"` prefixes?

src/utils.rs
```diff
@@ -10,6 +11,55 @@ pub unsafe fn debug_reset_static_data() {
     ffi::cairo_debug_reset_static_data()
 }
 
+pub fn status_to_result<T>(status: ffi::cairo_status_t, obj: T) -> Result<T, Error> {
+    match status {
+        ffi::STATUS_SUCCESS => Ok(obj),
+        ffi::STATUS_NO_MEMORY => Err(Error::NoMemory),
```
sdroege, line 17: Couldn't you use the `Into` impl on `Error` here?

src/utils.rs
```diff
@@ -10,6 +11,55 @@ pub unsafe fn debug_reset_static_data() {
     ffi::cairo_debug_reset_static_data()
 }
 
+pub fn status_to_result<T>(status: ffi::cairo_status_t, obj: T) -> Result<T, Error> {
```
sdroege, line 14: If you let it take a `FnOnce() -> T` closure instead then you could also use it in all the other places where an actual value is returned

src/utils.rs
bilelmoussaoui: Indeed, done

src/device.rs
bilelmoussaoui: hmm, I'm not sure how to do that. I just wanted to remove the Status usage here. What do you suggest to use?

src/utils.rs
bilelmoussaoui: I have found only one place where it can be useful which is the `add_outline` 
```
pub fn add_outline(
        &self,
        parent_id: i32,
        name: &str,
        link_attribs: &str,
        flags: PdfOutline,
    ) -> Result<i32, Error> {
        let name = CString::new(name).unwrap();
        let link_attribs = CString::new(link_attribs).unwrap();

        let res = unsafe {
            ffi::cairo_pdf_surface_add_outline(
                self.0.to_raw_none(),
                parent_id,
                name.as_ptr(),
                link_attribs.as_ptr(),
                flags.bits() as _,
            ) as _
        };

        self.status()?;
        Ok(res)
    }
```
The rest of the cases returns a `Result<(), Error>`. I can still change it to something that takes a FnOnce if that will make the code more idiomatic

src/device.rs
sdroege:
```rust
pub struct DeviceAcquireGuard<'a>(&'a Device);

impl<'a> Drop for DeviceAcquireGuard<'a> {
    fn drop(&mut self) { self.0.release(); }
}
```

And return that from `acquire()`

sdroege: (and derive `Debug`)

sdroege: (and make `release()` private)

src/utils.rs
sdroege: In that case leave it but remove the `obj` parameter and always make it return a `Result<(), Error>` :)

src/device.rs
bilelmoussaoui: Done, thanks for the pointer :)

bilelmoussaoui:
Thanks for the fast review & help:) It seems the CI failed due to network. Let me know if this needs any more changes

sdroege:
@GuillaumeGomez CI is green, please merge :)

GuillaumeGomez:
Thanks!

Issue #323, opened 2020-06-04
Segmentation fault in safe code when accessing raw data from a finished surface


bdonlan:
On cairo-rs 0.8.1, the following code results in a segmentation fault without using unsafe code:

```
use cairo::{ImageSurface, Format};

pub fn main() {
    let mut surf = ImageSurface::create(
        Format::ARgb32,
        1024,
        1024
    ).unwrap();

    surf.finish();

    let mut data = surf.get_data().unwrap();

    data[0] = 1;
}
```

sdroege:
I can reproduce the crash, thanks for reporting. Will investigate :)

sdroege:
Seems to be a bug in cairo. After `cairo_surface_finish()` it should in theory return `CAIRO_STATUS_SURFACE_FINISHED` from `cairo_surface_status()` but doesn't.

sdroege:
See https://gitlab.freedesktop.org/cairo/cairo/-/issues/406

psychon:
At least this fixes the immediate crash, but it is ugly (and of course this still needs an evil comment saying that cairo makes it really hard to check if some surface is finished):
```diff
diff --git a/src/image_surface.rs b/src/image_surface.rs
index ade61ab..c25cda6 100644
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -73,7 +73,7 @@ impl ImageSurface {
                 Status::Success => (),
                 status => return Err(BorrowError::from(status)),
             }
-            if ffi::cairo_image_surface_get_data(self.to_raw_none()).is_null() {
+            if ffi::cairo_image_surface_get_data(self.to_raw_none()).is_null() || is_finished(self) {
                 return Err(BorrowError::from(Status::SurfaceFinished));
             }
             Ok(ImageSurfaceData::new(self))
@@ -148,6 +148,12 @@ impl<'a> fmt::Display for ImageSurfaceData<'a> {
     }
 }
 
+fn is_finished(surface: &ImageSurface) -> bool {
+    use super::Context;
+    let ctxt = Context::new(surface);
+    ctxt.status() != Status::Success
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -172,4 +178,17 @@ mod tests {
         let result = ImageSurface::create_for_data(vec![0u8; 40 * 10], Format::ARgb32, 10, 10, 40);
         assert!(result.is_ok());
     }
+
+    #[test]
+    fn crash() {
[...]

PR #324, opened 2020-06-04
Work around crash when trying to access data of an image surface afte…

sdroege:
…r finish()

Fixes https://github.com/gtk-rs/cairo/issues/323

### DIFF
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -75,7 +75,8 @@ impl ImageSurface {
             if let Some(err) = status_to_result(status).err() {
                 return Err(BorrowError::from(err));
             }
-            if ffi::cairo_image_surface_get_data(self.to_raw_none()).is_null() {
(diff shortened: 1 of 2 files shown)


src/image_surface.rs
```diff
@@ -148,6 +149,14 @@ impl<'a> fmt::Display for ImageSurfaceData<'a> {
     }
 }
 
+// Workaround for cairo not having a direct way to check if the surface is finished.
+// See: https://gitlab.freedesktop.org/cairo/cairo/-/issues/406
+fn is_finished(surface: &ImageSurface) -> bool {
```
GuillaumeGomez, line 154: I don't what "finished" refers to here...

sdroege: `cairo_surface_finish()`, it's a concept of cairo surfaces

GuillaumeGomez: Ok, thanks!

PR #326, opened 2020-06-08
add a freetype feature

### LINKED ISSUE #325: `cairo_ft_*` functions are assumed to be available
I believe that right now, both the `-sys` crate and the main crate assume that Cairo is built with FreeType support, i.e. that the `cairo_ft_*` functions are defined. In the main crate, these functions are used here: https://github.com/gtk-rs/cairo/blob/master/src/font/font_face.rs#L95-L105

On Windows, cairo might not be built with FreeType, as in [the conda-forge cairo package](https://github.com/conda-forge/cairo-feedstock/). We recently ran into a build issue with librsvg that seems to stem from the violation of this assumption.

If this diagnosis is correct, and there's a desire to increase the build portability to include systems where Cairo is built without FreeType, I think that another build feature would have to be added, mapping to `CAIRO_HAS_FT_FONT`.

bilelmoussaoui:
cairo can be built without the ft_
fixes #325


sdroege:
:+1: 

@GuillaumeGomez CI is green also :)

GuillaumeGomez:
Thanks!

pkgw:
Oh hey, thanks for doing this!

PR #328, opened 2020-06-09
add a script feature

### LINKED ISSUE #259: Linker error if Cairo is configured with --disable-script
When building these Rust bindings, it currently fails during linking when Cairo is configured without the script backend (`--disable-script`).

For example compiling librsvg 2.45.6 produces these linker errors:
```
./.libs/librsvg-2.so: undefined reference to `cairo_script_set_mode'
./.libs/librsvg-2.so: undefined reference to `cairo_script_surface_create'
./.libs/librsvg-2.so: undefined reference to `cairo_script_from_recording_surface'
./.libs/librsvg-2.so: undefined reference to `cairo_script_surface_create_for_target'
./.libs/librsvg-2.so: undefined reference to `cairo_script_get_mode'
./.libs/librsvg-2.so: undefined reference to `cairo_script_write_comment'
collect2: error: ld returned 1 exit status
make[3]: *** [rsvg-convert] Error 1
make[3]: Leaving directory `/deps/svg'
make[2]: *** [install-recursive] Error 1
make[2]: Leaving directory `/deps/svg'
make[1]: *** [install] Error 2
make[1]: Leaving directory `/deps/svg'
make: *** [install-strip] Error 2
```

This can be fixed with [this patch](https://github.com/libvips/build-win64-mxe/blob/a3c7569426b213fa9993e6c4ff8a5eb1e4afbab2/8.8/patches/librsvg-2-fixes.patch#L209-L277), but instead of removing the code that causes these linker errors, it would be better if it discovers that Cairo has been built without a script backend.

bilelmoussaoui:
fixes #259


GuillaumeGomez:
Looks good to me as well, thanks!

PR #330, opened 2020-06-18
Add an ImageSurface::with_data method.

### LINKED ISSUE #329: Get data out of a referenced ImageSurface
`ImageSurface::get_data` returns an error if the surface is referenced, but I believe it should still be possible to copy the data out safely: since neither `ImageSurface` nor `Context` is `Sync`, there shouldn't be any concurrent modifications.

Of course, there's the question of how to return the data: you can't return `&[u8]`, for reasons discussed in #279. But would you accept a PR to add a method like this?
```rust
impl ImageSurface {
    pub fn copy_data_to(&self, buf: &mut [u8]) -> Result<(), BorrowError>;

    // Or, a more general version
   pub fn with_data<F: FnOnce(&[u8])>(&self, f: F) -> Result<(), BorrowError>;
}
```

jneem:
Fixes #329

### DIFF
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -64,6 +64,16 @@ impl ImageSurface {
         result
     }
 
+    /// Get the pixel data of the image surface, for direct inspection or modification.
+    ///
+    /// Unlike the C API, this method calls `flush` on the surface.
+    ///
+    /// Returns an error if [`finish`] was called on this surface, or if the surface is being
+    /// referenced by a [`Context`](struct.Context.html). See [`with_data`] if you need to access
+    /// the data on an image surface that is being referenced by a `Context`.
+    ///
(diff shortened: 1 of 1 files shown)


jneem:
So the CI is failing, but I don't understand why. Are API docs not allowed?

sdroege:
Something for the docs expert, @GuillaumeGomez? :)

I think it's complicated and manual docs have to be added to some other file, I'm not sure.

GuillaumeGomez:
Yes, because of license issues we have to store documentation elsewhere and it impacts other documentation too. So please add the doc [here](https://github.com/gtk-rs/lgpl-docs/blob/master/cairo/docs.md) instead.

GuillaumeGomez:
Thanks!

PR #331, opened 2020-06-29
Use system-deps crate instead of pkg-config directly

sdroege:


EPashkin:
Hm, I missed this in sys, why not use `eprinltn!` ? It stable since `1.19.0`

sdroege:
> Hm, I missed this in sys, why not use `eprinltn!` ? It stable since `1.19.0`

Where what?

sdroege:
@gdesmott looks like a bug in `system-deps`, it doesn't pull in `cairo-gobject` although the `use_glib` feature is given.

cairo-sys-rs/build.rs
```diff
@@ -17,67 +13,8 @@ fn main() {} // prevent linking libraries to avoid documentation failure
...
-        }
-        Err(Error::EnvNoPkgConfig(_)) | Err(Error::Command { .. }) => {
-            for lib_ in shared_libs.iter() {
-                println!("cargo:rustc-link-lib=dylib={}", lib_);
-            }
-            Ok(())
-        }
-        Err(err) => Err(err),
+    if let Err(s) = system_deps::Config::new().probe() {
+        let _ = writeln!(io::stderr(), "{}", s);
```
EPashkin, line 17: Here
this code snippet was in gir "from beginning"

sdroege: Ah, will fix once the `system-deps` issue is solved :)

cairo-sys-rs/Cargo.toml
```diff
@@ -10,6 +10,23 @@ keywords = ["cairo", "ffi", "gtk-rs", "gnome"]
...
+[package.metadata.system-deps."cairo"]
+name = "cairo"
+version = "1.12"
+
+[package.metadata.system-deps."cairo".feature-versions]
+v1_14 = "1.14"
+v1_16 = "1.16"
+
+[package.metadata.system-deps."cairo-gobject"]
+name = "cairo"
```
gdesmott, line 22: You need the actual lib name here, so  `cairo-gobject`, or you can just let it get it from the toml key name and remove this line.

sdroege: Oops. Thanks :)

PR #336, opened 2020-07-09
errors: derive Error from common enum traits

bilelmoussaoui:
Required to update librsvg to use the new cairo release


sdroege:
Thanks!

@GuillaumeGomez should probably also get a bugfix release to make librsvg happy :)

Issue #282, opened 2019-08-21
add FreeType fonts support


bilelmoussaoui:
Currently, cairo lacks those functions 

- [ ] `cairo_ft_font_face_create_for_ft_face ()`
- [ ] `cairo_ft_font_face_create_for_pattern ()`
- [ ] `cairo_ft_font_options_substitute ()`
- [ ] `cairo_ft_scaled_font_lock_face ()`
- [ ] `cairo_ft_scaled_font_unlock_face ()`
- [ ] `cairo_ft_font_face_get_synthesize ()`
- [ ] `cairo_ft_font_face_set_synthesize ()`
- [ ] `cairo_ft_font_face_unset_synthesize ()`

Documentation: https://www.cairographics.org/manual/cairo-FreeType-Fonts.html

Those allow using an external font with Cairo. There's also Win32 fonts & other missing stuff, see https://www.cairographics.org/manual/cairo-fonts.html.

GuillaumeGomez:
Interesting into adding them yourself? Otherwise you'll have to wait a bit.

bilelmoussaoui:
> Interesting into adding them yourself? Otherwise you'll have to wait a bit.

I will definitely add what I need & maybe the rest too.

GuillaumeGomez:
:+1:

derekdreery:
I'm doing some work where I need these functions now. @bilelmoussaoui I was wondering if you made any progress? The problem is that these functions take values from `freetype` so I would need to either import them or redefine them. Because they are just pointers, the temptation is to define them as opaque pointer types or `*mut c_void`.

PR #341, opened 2020-10-18
Add wrapper for `cairo_ft_font_face_create_for_ft_face`.

derekdreery:
I'm opening this PR for feedback initially. I need to be able to access `cairo_ft_font_face_create_for_ft_face` to be able to migrate [`druid`](https://github.com/linebender/druid) away from the toy text API. I've implemented the function here, but there are some things I'm unsure of:

 1. Types to use for the external types (`FT_Face` and `FcPattern`). I think this create should just define placeholder versions of these types. They are both treated as opaque pointers, so we can leave it up to the user to cast the pointer, knowing that it has come from freetype/fontconfig. That's how I've implemented it in this patch currently.
 2. Handling ownership of the font returned by `cairo_ft_font_face_create_for_ft_face`. The user supplied `FT_Face` must outlive the cairo font object, but I don't think it's possible to enforce this. Someone could write a library using both this and [`freetype`](https://docs.rs/freetype-rs/0.26.0/freetype/) and make the interface fully safe.
 3. A wrapper for `cairo_ft_font_face_create_for_pattern`. I know little of fontconfig and don't need this for my work.

This PR would at least partially address #282.

### DIFF
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -753,7 +753,8 @@ extern "C" {
         load_flags: c_int,
     ) -> *mut cairo_font_face_t;
     #[cfg(any(feature = "freetype", feature = "dox"))]
-    pub fn cairo_ft_font_face_create_for_pattern(pattern: *mut FcPattern) -> *mut cairo_font_face_t;
+    pub fn cairo_ft_font_face_create_for_pattern(pattern: *mut FcPattern)
+        -> *mut cairo_font_face_t;
     #[cfg(any(feature = "freetype", feature = "dox"))]
     pub fn cairo_ft_font_option_substitute(
         options: *const cairo_font_options_t,
--- a/cairo-sys-rs/src/lib.rs
+++ b/cairo-sys-rs/src/lib.rs
@@ -275,6 +275,11 @@ pub type cairo_read_func_t =
 pub type cairo_write_func_t =
     Option<unsafe extern "C" fn(*mut c_void, *mut c_uchar, c_uint) -> cairo_status_t>;
 
+#[cfg(any(feature = "freetype", feature = "dox"))]
+pub type FT_Face = *mut c_void;
+#[cfg(any(feature = "freetype", feature = "dox"))]
+pub type FcPattern = c_void;
+
 extern "C" {
     // CAIRO CONTEXT
     pub fn cairo_create(target: *mut cairo_surface_t) -> *mut cairo_t;
@@ -742,6 +747,22 @@ extern "C" {
     pub fn cairo_text_cluster_allocate(num_clusters: c_int) -> *mut TextCluster;
     pub fn cairo_text_cluster_free(clusters: *mut TextCluster);
 
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_font_face_create_for_ft_face(
+        face: FT_Face,
+        load_flags: c_int,
+    ) -> *mut cairo_font_face_t;
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_font_face_create_for_pattern(pattern: *mut FcPattern) -> *mut cairo_font_face_t;
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_font_option_substitute(
+        options: *const cairo_font_options_t,
+        pattern: *mut FcPattern,
+    );
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_scaled_font_lock_face(scaled_font: *mut cairo_scaled_font_t) -> FT_Face;
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_scaled_font_unlock_face(scaled_font: *mut cairo_scaled_font_t);
     #[cfg(any(feature = "freetype", feature = "dox"))]
     pub fn cairo_ft_font_face_get_synthesize(
         font_face: *mut cairo_font_face_t,
--- a/src/font/font_face.rs
+++ b/src/font/font_face.rs
@@ -8,7 +8,7 @@ use std::ptr;
 
 use enums::{FontSlant, FontType, FontWeight};
 
-#[cfg(any(feature = "use_freetype", feature = "dox"))]
+#[cfg(any(feature = "freetype", feature = "dox"))]
 use enums::FtSynthesize;
 
 use utils::status_to_result;
@@ -45,8 +45,8 @@ impl FontFace {
     }
 
     // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
-    #[cfg(any(feature = "use_freetype", feature = "dox"))]
-    pub unsafe fn create_from_ft(face: freetype::freetype::FT_Face) -> FontFace {
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub unsafe fn create_from_ft(face: freetype_crate::freetype::FT_Face) -> FontFace {
         let font_face = FontFace::from_raw_full(ffi::cairo_ft_font_face_create_for_ft_face(
             face as *mut _,
             0,
@@ -57,9 +57,9 @@ impl FontFace {
     }
 
     // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
-    #[cfg(any(feature = "use_freetype", feature = "dox"))]
+    #[cfg(any(feature = "freetype", feature = "dox"))]
     pub unsafe fn create_from_ft_with_flags(
-        face: freetype::freetype::FT_Face,
+        face: freetype_crate::freetype::FT_Face,
         load_flags: c_int,
     ) -> FontFace {
         let font_face = FontFace::from_raw_full(ffi::cairo_ft_font_face_create_for_ft_face(
@@ -123,17 +123,17 @@ impl FontFace {
         unsafe { ffi::cairo_font_face_get_reference_count(self.to_raw_none()) as usize }
     }
 
-    #[cfg(any(feature = "use_freetype", feature = "dox"))]
+    #[cfg(any(feature = "freetype", feature = "dox"))]
     pub fn get_synthesize(&self) -> FtSynthesize {
         unsafe { FtSynthesize::from(ffi::cairo_ft_font_face_get_synthesize(self.to_raw_none())) }
     }
 
-    #[cfg(any(feature = "use_freetype", feature = "dox"))]
+    #[cfg(any(feature = "freetype", feature = "dox"))]
     pub fn set_synthesize(&self, synth_flags: FtSynthesize) {
         unsafe { ffi::cairo_ft_font_face_set_synthesize(self.to_raw_none(), synth_flags.into()) }
     }
 
-    #[cfg(any(feature = "use_freetype", feature = "dox"))]
+    #[cfg(any(feature = "freetype", feature = "dox"))]
     pub fn unset_synthesize(&self, synth_flags: FtSynthesize) {
         unsafe { ffi::cairo_ft_font_face_unset_synthesize(self.to_raw_none(), synth_flags.into()) }
     }
--- a/src/font/font_face.rs
+++ b/src/font/font_face.rs
@@ -44,12 +44,6 @@ impl FontFace {
         font_face
     }
 
-    // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
-    #[cfg(any(feature = "freetype", feature = "dox"))]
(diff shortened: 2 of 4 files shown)


bilelmoussaoui:
I wonder if we can make use of https://docs.rs/yeslogic-fontconfig/0.1.1/fontconfig/, it provides the -sys part of the bindings as well.

derekdreery:
@bilelmoussaoui my thinking was that by just using a `*mut c_void` I'm not tying to any other libs, or requiring them as dependencies. If there is a standard `*-sys` library for fontconfig then maybe it might make sense to use that instead, but you could still end up with different types if you and a downstream crate used different versions of that library.

derekdreery:
There seem to be multiple competing `fontconfig-sys*` crates.

bilelmoussaoui:
> There seem to be multiple competing `fontconfig-sys*` crates.

I've only found two of them, the one used by the crate I mentioned above and a servo- one.

sdroege:
> @bilelmoussaoui my thinking was that by just using a `*mut c_void`

I didn't look at the PR yet, but if that's the case then I would only do the `-sys` part of the bindings as it would be all unsafe anyway.

Depending (optionally) on one of the freetype bindings seems like a good idea to me for the non-`sys` bindings.

derekdreery:
Ok I wasn't sure if you were willing to take a dep on freetype. I'll look at reworking the PR based on that.

sdroege:
As long as it's optional via some cargo feature flag this seems like the best way forward

derekdreery:
My concern with this is that different people will want to use different backends. For example, I'm planning to use font-kit to load fonts, but others may want to use [freetype-rs](https://github.com/PistonDevelopers/freetype-rs) or one of a number of other bindings. I feel like the ecosystem doesn't have a canonical freetype wrapper, and so the best option might be to just accept an opaque pointer (c_void style) and then leave it up to the user to select which freetype lib they want to use.

sdroege:
`font-kit` is not a freetype binding though, but an abstraction over freetype and the system libraries on Windows/macOS. And cairo would only work with the former or not?

If you want to go with raw pointers then let's keep it in the `-sys` bindings, there's not much advantage in having a function with raw pointers in the non-`sys` bindings here :)

derekdreery:
Ok that makes sense, higher level stuff can then be added by someone with more knowledge of what bindings would fit with the goals of gtk-rs.

sdroege:
I guess more by someone who knows how the Rust font ecosystem works :)

cairo-sys-rs/src/lib.rs
```diff
@@ -742,6 +747,22 @@ extern "C" {
...
+    pub fn cairo_ft_font_face_create_for_pattern(pattern: *mut FcPattern) -> *mut cairo_font_face_t;
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_font_option_substitute(
+        options: *const cairo_font_options_t,
+        pattern: *mut FcPattern,
+    );
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_scaled_font_lock_face(scaled_font: *mut cairo_scaled_font_t) -> FT_Face;
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_scaled_font_unlock_face(scaled_font: *mut cairo_scaled_font_t);
```
sdroege, line 765: The two functions above could get save bindings though, only the others with the `*mut c_void` don't make much sense

cairo-sys-rs/src/lib.rs
```diff
@@ -275,6 +275,11 @@ pub type cairo_read_func_t =
 pub type cairo_write_func_t =
     Option<unsafe extern "C" fn(*mut c_void, *mut c_uchar, c_uint) -> cairo_status_t>;
 
+#[cfg(any(feature = "freetype", feature = "dox"))]
+pub type FT_Face = *mut c_void;
+#[cfg(any(feature = "freetype", feature = "dox"))]
+pub type FcPattern = c_void;
```
sdroege, line 281: Why is one a pointer and the other not, but always used as a pointer below?

derekdreery:
I had another look at font-kit and it uses freetype under the hood, so i integrated with that. Users of other rust freetype crates can still drop down to the lower level methods in *-sys.

cairo-sys-rs/src/lib.rs
derekdreery: This is to match how the functions are defined in cairo - `FT_Face` is defined in freetype to be a pointer, but `FcPattern` in fontconfig is defined as an object. See https://cairographics.org/manual/cairo-FreeType-Fonts.html#cairo-ft-font-face-create-for-ft-face.

cairo-sys-rs/src/lib.rs
derekdreery: I'm sorry, what do you mean by "save bindings"?

cairo-sys-rs/src/lib.rs
derekdreery: We can define them differently in rust if you think it makes sense to.

cairo-sys-rs/src/lib.rs
sdroege: `s/save/safe/`, sorry. The non-`sys` ones.

cairo-sys-rs/src/lib.rs
sdroege: No, that explanation makes sense

cairo-sys-rs/src/lib.rs
derekdreery: Oh sorry, you mean "safe bindings". I will think about what this would look like.

sdroege:
> I had another look at font-kit and it uses freetype under the hood

From the readme on the website it sounds like freetype is only one of its backends, and that the backend depends on the platform. Is this going to work properly on Windows/macOS?

cairo-sys-rs/src/lib.rs
sdroege: Looks like something like a `MutexGuard`-style guard type can be used here

cairo-sys-rs/src/lib.rs
```diff
@@ -742,6 +747,22 @@ extern "C" {
...
+        face: FT_Face,
+        load_flags: c_int,
+    ) -> *mut cairo_font_face_t;
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_font_face_create_for_pattern(pattern: *mut FcPattern) -> *mut cairo_font_face_t;
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub fn cairo_ft_font_option_substitute(
+        options: *const cairo_font_options_t,
+        pattern: *mut FcPattern,
+    );
```
sdroege, line 761: And if we go for the `freetype` crate then the others can also be added to the non-`sys` bindings :)

cairo-sys-rs/src/lib.rs
derekdreery: There might be a problem with defining safe methods here: the functions expect a font face object created from freetype, but this isn't enforced in the type system. If a non-freetype font is passed to these functions I don't know what will happen.

derekdreery: We could track whether the `FontFace` in rust was created from freetype, possibly using zero-sized types to make it zero cost, but it would add complexity.

sdroege: Can that be checked at runtime and be handled via a `panic!`? Otherwise mark the functions `unsafe`

derekdreery:
> > I had another look at font-kit and it uses freetype under the hood
>
> From the readme on the website it sounds like freetype is only one of its backends, and that the backend depends on the platform. Is this going to work properly on Windows/macOS?

This patch won't mention font kit at all - I will get the raw freetype handle out of font kit and pass it to a function that expects a `freetype::freetype::FT_Face`. Freetype is useable on macos/windows, but usually users from these platforms will use the platform libraries for font handling. In any case, users of cairo can choose whether to include the freetype methods or not using the feature flag.

cairo-sys-rs/src/lib.rs
derekdreery: Yes there is the possibility of enforcing the locking rules at compile time using Rust's ownership model. I would need to think through all the details of this as the ownership models of freetype and cairo are different and interact in potentially complex ways. It might be safer to let the user handle this unless we are sure we can provide a safe abstraction.

cairo-sys-rs/src/lib.rs
derekdreery: Potentially, yes.

derekdreery:
For some reason my computer isn't always showing me the latest messages on this thread, so apologies if I seem to not have noticed messages you have sent - I will once I press F5. :)

cairo-sys-rs/src/lib.rs
sdroege: Let's worry about that some other time then

cairo-sys-rs/src/lib.rs
sdroege: Want to give it a try? Otherwise let's get this in, I guess :)

derekdreery: I'll have a go.

cairo-sys-rs/src/lib.rs
derekdreery: I tried adding in a marker value for tracking font type at runtime, but it doesn't work with `glib_wrapper!`, so some work would be required in `glib-rs` to make it possible. I think it's probably best to leave it for now as you say.

cairo-sys-rs/src/lib.rs
derekdreery: Edit: Actually lets finish this PR without. They can always be added later, and it keeps the complexity of this PR lower.

derekdreery:
My remaining concern is that I had to change the name of the feature `freetype` so as not to clash with the `freetype` crate. This means that users of the crate that set the feature manually will need to change the name of the feature in their code. Is there a changelog I should put this information? Or are users comfortable with these kinds of changes at the moment, given its a new library in development?

sdroege:
An alternative to that would be renaming in `Cargo.toml` like I do in [async-tungstenite](https://github.com/sdroege/async-tungstenite/blob/master/Cargo.toml#L43-L46). I think that would be nicer as then we could keep the `freetype` feature, and from a user point of view the renaming is transparent.

derekdreery:
Ok I will make that change.

src/font/font_face.rs
```diff
@@ -44,6 +44,33 @@ impl FontFace {
         font_face
     }
 
+    // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub unsafe fn create_from_ft(face: freetype::freetype::FT_Face) -> FontFace {
```
sdroege, line 49:
```suggestion
    pub unsafe fn create_from_ft(face: freetype_crate::freetype::FT_Face) -> FontFace {
```

src/font/font_face.rs
```diff
@@ -44,6 +44,33 @@ impl FontFace {
...
+        ));
+        let status = unsafe { ffi::cairo_font_face_status(font_face.to_raw_none()) };
+        status_to_result(status).expect("Failed to create a FontFace");
+        font_face
+    }
+
+    // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
+    #[cfg(any(feature = "freetype", feature = "dox"))]
+    pub unsafe fn create_from_ft_with_flags(
+        face: freetype::freetype::FT_Face,
```
sdroege, line 62:
```suggestion
        face: freetype_crate::freetype::FT_Face,
```

derekdreery: Oops. I've force-pushed these fixes.

sdroege (approved):

derekdreery:
[looking at the source](https://github.com/freedesktop/cairo/blob/1676935768b888b3e4038b6dddb4f6362c137345/src/cairo-ft-font.c#L3867) it seems that the font type is checked, so we can assume that passing the wrong type of font in to these functions does not result in UB.

src/font/font_face.rs
```diff
@@ -44,6 +44,33 @@ impl FontFace {
         font_face
     }
 
+    // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
```
GuillaumeGomez, line 47: Maybe better to use a doc comment here, no?

GuillaumeGomez:
It looks good to me (even though it seems a bit weird).

src/font/font_face.rs
derekdreery: I will have a look at this tomorrow.

derekdreery:
For me, this is ready to be merged. I will add some docs to the docs repo separately.

derekdreery:
CI failure is spurious

PR #343, opened 2020-10-23
Re-export cairo-sys as `ffi`.

### LINKED ISSUE #342: Consider re-exporting `cairo-sys-rs` under `ffi`.
I ran into a situation recently where for reasons I don't fully understand rust was trying to link to libcairo twice: once from me using `cairo-sys-fs` directly, and once from it being a dep of `cairo`. If you re-exported the "C" interface, then this wouldn't be necessary. `ffi` seems to be a standard name (used in `std`) and unlikely to clash with anything else in the crate.

richard-uk1:
Closes #342 

2 questions:

 - Do you prefer with `doc(inline)` or without? Run `cargo doc` with and without to see the difference.
 - What other crates should I do this for?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,7 @@
 //!
 //! * **win32-surface** - Microsoft Windows surface support
 
-extern crate cairo_sys as ffi;
+pub extern crate cairo_sys as ffi;
 extern crate libc;
 extern crate thiserror;
 


sdroege:
> Do you prefer with `doc(inline)` or without? Run `cargo doc` with and without to see the difference.

@GuillaumeGomez ?

> What other crates should I do this for?

I guess this should be done for glib, gio, pango, pangocairo, atk, gdk-pixbuf, gdk, gtk, graphene, gsk4, gdk4, gtk4. If you don't want to do it for all of them that's ok, we can do the others.

sdroege:
> I guess this should be done for glib, gio, pango, pangocairo, atk, gdk-pixbuf, gdk, gtk, graphene, gsk4, gdk4, gtk4. If you don't want to do it for all of them that's ok, we can do the others.

In case of glib there's glib_sys and gobject_sys. Which should probably become `glib::ffi` and `glib::gobject_ffi` I guess?

GuillaumeGomez:
Without `doc(inline)` preferably.

derekdreery:
I've removed `doc(inline)`. As a bonus, it makes the diff very small :D.

As for the other crates, I would maybe prefer if you made the changes, because you will be more aware of things like the `gobject_ffi` naming issue. But if you want you can ask me to change some or all of them and I will do the corresponding PRs. Let me know on this issue which ones you would like me to change, if any.

GuillaumeGomez:
Hey @derekdreery  ! We migrated the cairo crate into the gtk-rs/gtk-rs repository. Could you please reopen your PR there? :)

PR #345, opened 2020-10-25
Add an unsafe variant of ImageSurface::create_for_data

### LINKED ISSUE #344: Why ImageSurface::create_for_data requires static lifetime?
I'm writing a simple code which uses `sdl2` and `cairo` together.  I want to update SDL2 `Texture` with cairo drawing, but there is a lifetime issue that prevents it.

The following is a snippet that cause the issue:

```rust
    let mut texture = texture_creator.create_texture_streaming(PixelFormatEnum::ARGB8888, 120, 120)?;
    texture.with_lock(None, |pixels, stride| {
        let surface = ImageSurface::create_for_data(pixels, Format::ARgb32, 120, 120, stride as i32).unwrap();
        let cr = Context::new(&surface);

        // Do some drawing with cr
    });
```
The argument of the closure, `pixels`, is alive within the scope of entire lambda, since we have no access to the pixels once the texture is unlocked.  However, `ImageSurface::create_for_data` requires a static lifetime for the `data` argument (`data: D`, `D: AsMut<[u8]> + 'static`).  Therefore, we get a lifetime error at the creation of the surface.

There are some workarounds: maybe we can write an unsafe code so that the lifetime check is ignored (though I don't know how to do it).  Otherwise, we can create a surface and do some drawing to it, and then copy the pixels to that of SDL2 texture.  But it is less efficient since it requires to copy the pixel array twice.

So my questions are:
* Why `data: D` requires a static lifetime?
* What is the best workaround?  Are there any other better way?

TonalidadeHidrica:
Closes #344 

### Discussion:
- There are several candidates for how to receive `data` as argument:
  - Receive only the beginning pointer (the current one)
  - Receive the beginning pointer and the length of slice
  - Receive `Range<*mut u8>` (in favor of [`slice::as_mut_ptr_range`](https://doc.rust-lang.org/std/primitive.slice.html#method.as_mut_ptr_range))

### Foodnote:
Any suggestion, including grammatical mistakes of unnatural English in the documentation, is welcome.

### DIFF
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -31,12 +31,11 @@ impl ImageSurface {
         }
     }
 
-    /**
-     * Creates an image surface for the provided pixel data.
-     * The pointer `data` is the beginning of the underlying slice,
-     * and at least `width * stride` succeeding bytes should be allocated.
-     * Additionally, `data` must live longer than any reference to the returned surface.
-     */
+    /// Creates an image surface for the provided pixel data.
+    /// - The pointer `data` is the beginning of the underlying slice,
+    ///   and at least `width * stride` succeeding bytes should be allocated.
+    /// - `data` must live longer than any reference to the returned surface.
+    /// - You have to free `data` by yourself.
     pub unsafe fn create_for_data_unsafe(
         data: *mut u8,
         format: Format,
@@ -44,8 +43,13 @@ impl ImageSurface {
         height: i32,
         stride: i32,
     ) -> Result<ImageSurface, Error> {
-        let data = std::slice::from_raw_parts_mut(data, (stride * height) as usize);
-        Self::create_for_data(data, format, width, height, stride)
+        ImageSurface::from_raw_full(ffi::cairo_image_surface_create_for_data(
+            data,
+            format.into(),
+            width,
+            height,
+            stride,
+        ))
     }
 
     pub fn create_for_data<D: AsMut<[u8]> + 'static>(
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -31,6 +31,7 @@ impl ImageSurface {
         }
     }
 
+    // rustdoc-stripper-ignore-next
     /// Creates an image surface for the provided pixel data.
     /// - The pointer `data` is the beginning of the underlying slice,
     ///   and at least `width * stride` succeeding bytes should be allocated.
--- a/src/image_surface.rs
+++ b/src/image_surface.rs
@@ -31,6 +31,23 @@ impl ImageSurface {
         }
     }
 
+    /**
+     * Creates an image surface for the provided pixel data.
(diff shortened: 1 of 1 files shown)


src/image_surface.rs
```diff
@@ -31,6 +31,23 @@ impl ImageSurface {
...
+     */
+    pub unsafe fn create_for_data_unsafe(
+        data: *mut u8,
+        format: Format,
+        width: i32,
+        height: i32,
+        stride: i32,
+    ) -> Result<ImageSurface, Error> {
+        let data = std::slice::from_raw_parts_mut(data, (stride * height) as usize);
+        Self::create_for_data(data, format, width, height, stride)
```
sdroege, line 48: You can have this one here a bit easier and without the potential of UB by going via a slice and all the complications with the user data:

```rust
        unsafe {
            ImageSurface::from_raw_full(ffi::cairo_image_surface_create_for_data(
                data,
                format.into(),
                width,
                height,
                stride,
            ))
        }
```

src/image_surface.rs
```diff
@@ -31,6 +31,23 @@ impl ImageSurface {
         }
     }
 
+    /**
+     * Creates an image surface for the provided pixel data.
+     * The pointer `data` is the beginning of the underlying slice,
+     * and at least `width * stride` succeeding bytes should be allocated.
+     * Additionally, `data` must live longer than any reference to the returned surface.
+     */
```
sdroege, line 39: Please make these normal documentation comments with `///`

TonalidadeHidrica: I confused with Java-like comments :P

src/image_surface.rs
TonalidadeHidrica: So we don't have to store the given pointer as user data, but just redirect the arguments to the native function?

sdroege: Exactly

TonalidadeHidrica:
Fixed!

sdroege (approved):

TonalidadeHidrica:
@sdroege Can I ignore the failing lints?  At a glance I have no idea how to fix it.

sdroege:
Appveyor you mean? They have some kind of network problem

sdroege:
Gah, but there's also some problem with the docs here. @GuillaumeGomez how is the docs stuff supposed to work, does this have to be added to the `lgpl-docs` repo?

GuillaumeGomez:
No, you need to add `// rustdoc-stripper-ignore-next` like shown [here](https://github.com/guillaumegomez/rustdoc-stripper#ignore-doc-comment) at the beginning of each added doc-comment.

TonalidadeHidrica:
@GuillaumeGomez Like this?

GuillaumeGomez:
Yes, like this! :)

